PR #5, opened 2015-01-08
support a blocked list, allow metadata in trello-yaml-card

bmwiedemann:


cornelius:
I did some additional cleanups, added tests, and merged the patch. Could you check, if it still works for you?

PR #8, opened 2015-01-22
Refactoring of trollolo

kalabiyau:
What is done with that refactoring:

* Moved to ruby-trello gem for trello api interaction in OO manner
* Simplified burndown data collection - directly from scrum_board object
* Introduced new settings to define name of the columns (still WIP though)
* Column, Card introduced/refactored
* All webmocks unified to support ruby trello as latter uses board -> list -> cards flow to get to the cards
* gemspec simplified - dev deps moved to Gemfile
* fast lane introduced - special type of a card which pops in the middle of a Sprint and should be done. Not having estimation though, therefore we have it plotted as amount of fast lane cards left in sprint

Not done yet

* Move column names settings to the board which will allow us to have per board columns names settings
* Other parts of app still requires refactoring

Please check this branch against your boards if that still working as expected.


kalabiyau:
Also contains fix for travisCI

lib/settings.rb
```diff
@@ -26,8 +27,12 @@ def initialize config_file_path
       @config = YAML.load_file(config_file_path)
 
       if @config
-        @developer_public_key = @config["developer_public_key"]
-        @member_token = @config["member_token"]
+        @developer_public_key   = @config["developer_public_key"]
+        @member_token           = @config["member_token"]
+        @not_done_columns       = @config["not_done_columns"].freeze
```
cornelius: I would suggest to do

```ruby
        @not_done_columns       = @config["not_done_columns"].freeze || ["Sprint Backlog", "Doing"]
```

here, so we have the old behavior by default. Then it also works with our data out of the box.

lib/result.rb
```diff
@@ -0,0 +1,12 @@
+class Result
```
cornelius: The class is a helper class for `BurndownData`. So from my point of view it fits there quite well. If we want to use it as a stand-alone class we should find a better name. `Result` is too generic. I was thinking of `BurndownDataResult`, but then we can also just leave it inside the `BurndownData` class.

spec/burndown_data_spec.rb
```diff
@@ -5,183 +5,64 @@
...
-          "done" => 6
-        },
-        "tasks_extra" => {
-          "done" => 8
-        }
-      }
-
-      expect(@burndown.to_hash).to eq(expected_hash)
-    end
-  end
```
cornelius: Why did you remove this test?

spec/burndown_data_spec.rb
```diff
@@ -5,183 +5,64 @@
...
-          
-      stub_request(:any,list_url_match).to_return(:status => 200,
-        :body => load_test_file("lists.json"), :headers => {})
 
-      expect( @burndown.fetch_done_list_id ).to eq "5319bf088cdf9cd82be336b0"
-    end
-    
-    it "raises exception if it does not find done column" do
-      expect{ @burndown.fetch_done_list_id }.to raise_error
-    end
```
cornelius: It would be good to keep the tests which check the error cases. Or did I just miss them being at another place?

lib/card.rb
```diff
@@ -17,32 +17,55 @@
 
 class Card
 
-  attr_accessor :meta, :sp, :tasks, :tasks_done, :title, :description
+  # Assuming we have card titles as foolows '(8) This is the card name'
```
cornelius: Typo: s/foolows/follows/

cornelius:
Thanks for the patch. I like the refactoring of the Spaghetti methods to the `Column` and `ScrumBoard` classes. That makes the code much more easy to read and provides more flexibility for the future.

Would it make sense to add some tests for these classes to make expectations clear, have some protection against regressions, and make it more easy to do future refactorings?

It might also make sense to split up the API a little bit as especially the `ScrumBoard` class is quite big now. But this is nothing which needs to be done right now.

The history of this pull request is mixed up. Commit messages and changes not always fit together and some commits have several unrelated changes. I would suggest to either squash this into one big commit or do a few logical pieces.

Oh and many thanks for fixing Travis :clap:. This was bothering me for a while, but I never got around to actually fix it.

spec/burndown_data_spec.rb
kalabiyau: Added to scrum_board_spec in latest squash

kalabiyau:
@cornelius squashed, fixed and ready to be re-reviewed. Sorry for `force`, but as agreed - that batch of commits been squashed.

cornelius:
Reviewed, tested, and merged. Thanks for all the work. :sparkles:

PR #11, opened 2015-01-29
argparse to plotter, add no-tasks switch to helper

kalabiyau:
As we don't need to track tasks in the team, I've introduced that parameter which can disable plotting tasks line in the chart.

So no changes required from existing users. Also argparse allowed us to nicely format and handle argv processing for plotter, so we don't need to care about them manually anymore.


cornelius:
The command `trollolo burndown --plot` now fails with an error. Could you check that?

cornelius:
It also would be great if you could rebase based on the latest changes in the python script. Sorry for the additional work.

PR #12, opened 2015-02-01
No tasks switch

kalabiyau:
* Rework of previous commit about the same feature
 
As we don't need to track tasks in the team, I've introduced that parameter which can disable plotting tasks line in the chart.

So no changes required from existing users. Also argparse allowed us to nicely format and handle argv processing for plotter, so we don't need to care about them manually anymore.

There were error mentioned by you @cornelius. Could you please drop a line how the erorr looks like and if that is still the case for this PR

### DIFF
--- a/lib/burndown_chart.rb
+++ b/lib/burndown_chart.rb
@@ -14,7 +14,6 @@
 #
 #  To contact SUSE about this file by physical or electronic mail,
 #  you may find current contact information at www.suse.com
-
 class BurndownChart
 
   attr_accessor :data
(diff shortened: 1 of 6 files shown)


lib/burndown_chart.rb
```diff
@@ -114,9 +114,10 @@ def setup(burndown_dir, board_id)
     write_data File.join(burndown_dir, burndown_data_filename)
   end
 
-  def self.plot(sprint_number, working_dir=".")
+  def self.plot(sprint_number, options)
     plot_helper = File.expand_path("../../scripts/create_burndown.py", __FILE__ )
-    system "python #{plot_helper} #{sprint_number} #{working_dir}"
+    cli_switches = options.keys.map{|k| k.dup.prepend('--')}
```
cornelius: I would prefer to make this explicit and check the options, which are passed on. It might also be good to add a test for this method.

With this change it doesn't work, because `trollolo` accepts an `--output` option, but the python rendering script expects a `--working-dir`.

cornelius: This also breaks, when called as `trollolo burndown --plot` as the update method still expects the old `plot` signature.

kalabiyau:
@cornelius could you please give it another review. Most important is the change to python plotter

cornelius:
I merged this now and added some additional tests checking the output of the Python script.

PR #16, opened 2015-04-16
Major refactoring of create_burndown.py

aosthof:
As the script became pretty unmaintainable over time, I've refactored the code in an object oriented way. Please review my changes and give feedback.


scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
-  y_text = - tasks_done_extra / 2
-  plt.text(0.4, y_text, str(int(tasks_done_extra)) + " extra",
-	   rotation='vertical', verticalalignment='center', color='green'
-	  )
-  plt.text(0.7, y_text, "tasks done",
-	   rotation='vertical', verticalalignment='center', color='green'
-	  )
 
+# Parse arguments and/or print help
+def argParser():
```
cornelius: I would name the function `parseCommandLine()`, this would make it clearer what happens.

The comment should not repeat what already is in the function name, but add information which is not explicitly there. In this case this would be the API documentation, i.e. how to use the function, when to call it, what it returns.

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+    print args
+
+  return args
+
+
+
+class BurndownData:
+  "Store burndown data parsed from YAML file"
+
+  # Constructor
```
cornelius: This comment is redundant. In Python it's clear that `__init__` is the constructor.

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+  def __init__(self, args):
+    burndown = self.openYAML(args.sprint)
+    self.getSprintData(burndown)
+
+  # Open burndown yaml file
+  def openYAML(self, sprint_number):
```
cornelius: I would not only open the YAML in this function. It would be a more logical unit to read the data from the YAML file and store it in some internal data structure. So opening the YAML, readings it's content, and storing it in member variables, could be done by a function named `readYAML`.

Storing in member variables would be preferable over returning it as return value, as the data is used in subsequent methods.

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+    self.total_fast_lane = []
+    self.max_story_points = 0
+  
+    for day in burndown["days"]:
+      self.days.append(current_day)
+      self.open_story_points.append(day["story_points"]["open"])
+      self.total_story_points.append(day["story_points"]["total"])
+      self.open_tasks.append(day["tasks"]["open"])
+      self.total_tasks.append(day["tasks"]["total"])
+      self.max_story_points = max(self.max_story_points, day["story_points"]["total"])
```
cornelius: I would separate processing data and reading it from the YAML. Calculating the maximum story points is different from reading the raw data. So I would move all the code which directly just reads data to the `readYAML` function and have dedicated functions for calculating the data needed for the plot.

In this case it might even make sense to have a separate function `maxStoryPoints`, which calculates the maximum value and returns it, so that the calling code can use it or store it.

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+    
+      # TODO: refactor to not to draw 0 extra Story Points Line if one task
+      # in extra card been done. Example http://imagebin.suse.de/1785/img
```
cornelius: Is this comment still relevant?

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+    # If there are already done tasks at the beginning of a sprint
+    if burndown["days"][0].has_key("tasks_extra"):
+       self.bonus_tasks_day_one = burndown["days"][0]["tasks_extra"]["done"]
+    
+    # Add a day at the beginning of the extra days, so the curve starts at zero
+    if len(self.extra_days) > 0 and not self.bonus_tasks_day_one:
+      self.extra_days = [self.extra_days[0] - 1] + self.extra_days
+      self.bonus_story_points_done = [0] + self.bonus_story_points_done
+      self.bonus_tasks_done = [0] + self.bonus_tasks_done
+      self.extra_day = 1
```
cornelius: This is another example for calculation, which could be extracted to a separate function. This would make the code clearer, remove the need for a comment, and make it possible to test it separately in a unit test.

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+
+    return
+
+
+class PlotConfig:
```
cornelius: I'm not sure `PlotConfig` is the best name for this class. It represents the overall plot, not only the configuration. It actually plots everything but the graphs.

Maybe just call it `Plot`?

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+    self.setYAxisLimits(data.max_story_points, data.bonus_tasks_done, data.bonus_story_points_done)
+    self.createFigure(figure_width, figure_height)
+    self.setTitle(data.sprint_number, title_fontsize)
+    self.setXAxisLabel(label_x_axis)
+    self.plotDiagonal(1, data.total_days, data.max_story_points, 0, color_diagonal)
+    self.plotWaterLine(0, data.total_days+1, 0, 0, color_waterline, linestyle_waterline)
+    self.plotWeekendLines(data.weekend_lines, self.ymin+1, self.ymax-1, color_weekend, linestyle_weekend)
+    self.setScaleFactor(data.total_tasks[0], data.max_story_points)
+    self.setXAxisLimits(0, data.total_days+1)
+    self.setXAxisTicks(data.days)
```
cornelius: Doing all this in the constructor is not very flexible and doesn't make it very clear what is happening. I would extract the code, which explicitly sets what the user sees from the code which does the implicit setup. So from a user of this class it would be clearer to me, if it would look like something like this:

```python
plot = Plot(data)

plot.setXKCD()
plot.setTitle("Sprint " + data.sprint_number)
plot.drawDiagonal("grey")
plot.drawWaterline("blue")
plot.drawWeekendLines("grey")
```

and then have code drawing the graphs similar to something like:

```python
graph_story_points = Graph("Story Points", data.storyPoints())
plot.drawGraph(graph_story_points, "black")

graph_tasks = Graph("Tasks", data.tasks())
plot.drawGraph(graph_tasks, "green")
```

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+
+  # Plot the chart in XKCD style
+  def setXKCD(self):
+    plt.xkcd()
+    return
+
+  # Calculate minimum and maximum values for the y axis
+  def setYAxisLimits(self, max_story_points, bonus_tasks_done, bonus_story_points_done):
```
cornelius: Maybe better call the function `calculateYRange` and let it return the calculated values.

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+    if not args.no_tasks:
+      self.setTasksConfig(label_tasks, color_tasks, config.ymin, config.ymax, config.scalefactor)
+      self.plotTasks(data.days, data.extra_days, data.open_tasks, data.bonus_tasks_done, linestyle_tasks, linewidth_tasks)
+      self.plotNewTasks(data.total_tasks, width_tasks_bar, color_tasks_bar)
+      self.plotTasksDiff(data.total_tasks, data.open_tasks, data.bonus_tasks_day_one, data.bonus_tasks_done, data.max_story_points, data.days, config.scalefactor, color_tasks_diff)
+
+    if args.with_fast_lane:
+      self.setFastLaneConfig(label_fast_lane, color_fast_lane, config.ymin, config.ymax, config.fig)
+      self.plotFastLane(data.x_fast_lane, data.y_fast_lane, linestyle_fast_lane, linewidth_fast_lane)
+      self.plotNewFastLane(data.total_fast_lane, width_fast_lane_bar, color_fast_lane_bar)
```
cornelius: I don't think that most of this logic belongs into the `Graph` class, let alone in its constructor. The argument parsing definitely should be done by the caller of the class, not the class itself. The class and the control flow would be much simpler, if it would get less parameters. So for example instead of passing `args` the caller could do something like:

```python
plot = Plot()
...

if not args.no_tasks:
  graph = Graph("Tasks",data.tasks)
  plot.drawGraph(graph)
```

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
 # Save the burndown chart
-plt.savefig('burndown-' + args.sprint + '.png',bbox_inches='tight')
-if not args.no_head:
-  plt.show()
+def saveBurndown(args):
+  plt.savefig('burndown-' + args.sprint + '.png',bbox_inches='tight')
+  if not args.no_head:
+    plt.show()
+
+  return
```
cornelius: This function belongs to the `Plot` class and it probably should better be called `saveImage` or something like that.

scripts/create_burndown.py
```diff
@@ -17,211 +17,402 @@
...
+bd = BurndownData(args)
+
+# Configure plot parameters
+pc = PlotConfig(bd)
+
+# Plot all graphs
+graph = Graph(bd, args, pc)
+
+# Save the burndown chart
+saveBurndown(args)
```
cornelius: In this part the actual control flow of creating the burndown chart takes place. It would be clearer, if it would more work like shown in the examples above.

So instead of calling one function after the other and forwarding return values as parameters it could create the relevant objects and call their methods to get the data, build up the Plot, and draw the individual elements.

cornelius:
We should make sure that the integration tests pass before we merge this code, so we have a reliable reference to check against, when changing the code.

Unit tests for the python code would be nice, but I would be ok with not doing them right now, as we have none right now anyway.

It might also be nice to split up the code and put each class in a separate file. This would make it much clearer, what belongs where.

The comments are a bit excessive and the very often just repeat what the code or the function name already says or does. I would rather use speaking function names and limit comments to API documentation and explanation of algorithms which need additional context beyond the code itself.

cornelius:
The commit history needs to be cleaned up. We should squash the branch into a single commit before merging.

cornelius:
There is a bug in the positioning of the bars indicating new tasks. They are moved one day to the right. You can see that when running the integration test for sprint 23.

cornelius:
When there are extra tasks done at the beginning of the sprint, the label and arrow indicating them are at the wrong position and it says "-6 extra tasks done". See integration test for sprint 31.

cornelius:
When the data is not complete yet, the x axis doesn't show labels for all days. You can see that in integration test for sprint 36.

aosthof:
The bug in the positioning of the bars indicating new tasks is fixed with commit 7ea58152c3a74febb81460c57aa4b4e1134b0309

aosthof:
The bug in the "extra tasks done" label is fixed with commit 2fc5e7b0862982886070cef91ebc755e29943904

aosthof:
The bug regarding incomplete x axis labels (days) is fixed with commit bbea829ff6bc54feb2fd19648fb00b48ef3467c6

cornelius:
The bars for new tasks are still shifted. Compare the integration test for sprint 23.

PR #19, opened 2015-09-24
Auto create new sprint

felixsch:
Creates a new sprint file if a new sprint column is detected online.


felixsch:
One test is failing due the change in the regex pattern used to determine the done columns. Because the general finding method relies on the changed regex ( look: https://github.com/openSUSE/trollolo/blob/master/lib/scrum_board.rb#L20 ). I suggest to change the test accordingly and update the manpage.

kalabiyau:
@digitaltom could you please help @felixsch with that PR?

felixsch:
I removed the by_max testing (therefore no specific regex is required) and returning the most left aligned column (relying on sorting from trello). A new sprint is now determined by the column (list) id. If the id changes a new sprint is created (even if the old sprint is resurrected).

cornelius:
Why do you prefer the column id over the sprint number parsed from the title of the done column? And why do do you prefer the left-most column over the one with the highest sprint number?

Using the sprint numbers has the advantage that it's more robust, because you can rearrange columns, or even recreate them.

felixsch:
This is because everyone can configure their own regex which captures the columns. In current situation the default regex (which matches the column) even does not capture sprint numbers. So I decided to stick by the columnid which is unique to a column even if the name changes.

cornelius:
I think your idea to go with the column is actually the right one. It makes it harder to replace the done column by another one on Trello, but on the other hand it is less susceptible to changes in the column names or other changes, which are done on the board. As it works transparently and adds a column id, when no one is there, and otherwise checks for the new column, it works nicely.

The code to check if a new column is there is currently all contained in one method and duplicates some functionality, which is alrerady implemented in other places, e.g. finding the latest sprint file, getting the data from the board. It would be better to restructure this a bit, so that the checks are done within the program flow of the burndown generation and is reusing the other methods in order to eliminate code duplication and minimize the number of HTTP request we do to the Trello API.

Ana06:
@felixsch @cornelius @kalabiyau @digitaltom What is the state of this? Are there plans to finish it or can we close it? :confused:

Ana06:
I guess I can close this. If I am wrong, please just reopen it. :smile:

PR #20, opened 2015-09-24
Add push to api endpoint option

felixsch:
Adds an option to send a JSON encoded POST request to a specified URL on burndown command in addition to the normal behaviour.


lib/burndown_chart.rb
```diff
@@ -14,6 +14,11 @@
 #
 #  To contact SUSE about this file by physical or electronic mail,
 #  you may find current contact information at www.suse.com
+
+
+require 'net/http'
```
kalabiyau: you probably already have them. Check if will work without those requires

lib/burndown_data.rb
```diff
@@ -44,6 +44,7 @@ def initialize(settings)
     @extra_story_points = Result.new
     @extra_tasks        = Result.new
     @fast_lane_cards    = Result.new
+    @date_time          = Date.today
```
kalabiyau: better use time here
Time.now

lib/burndown_data.rb
```diff
@@ -75,6 +76,26 @@ def to_hash
     end
   end
 
+  def to_api
```
kalabiyau: I bet that is rather to_json

kalabiyau:
@felixsch LGTM
@cornelius could you please review that one?

spec/unit/burndown_chart_spec.rb
```diff
@@ -302,6 +316,33 @@
...
+        expect(uri).to receive(:path).and_return url
+        expect(uri).to receive(:hostname).and_return hostname
+        expect(uri).to receive(:port).and_return 80
+
+        post = double
+        expect(post).to receive(:body=).with(burndown_data.to_json)
+
+        expect(URI).to receive(:parse).with(url).and_return(uri)
+        expect(Net::HTTP::Post).to receive(:new).with(url, header).and_return(post)
+        expect(Net::HTTP).to receive(:start).with(hostname, 80)
```
cornelius: I would use webmock here. That gives simpler and more robust code.

lib/burndown_chart.rb
```diff
@@ -106,6 +106,30 @@ def write_data filename
     end
   end
 
+
+  # Writes a POST request to url
+  def write_data_to_api(url, burndown_data)
```
cornelius: If the option is called `push_to_api` the method could also be called this way.

lib/burndown_data.rb
```diff
@@ -75,6 +76,26 @@ def to_hash
...
+      total_ep: extra_story_points.total.round,
+      open_ep:  extra_story_points.open.round,
+
+      total_tasks: tasks.total,
+      open_tasks:  tasks.open,
+
+      total_etasks: extra_tasks.total,
+      open_etasks:  extra_tasks.open
+    }.to_json
+  end
```
cornelius: There should be a unit test for the `to_json` card.

lib/burndown_data.rb
```diff
@@ -75,6 +76,26 @@ def to_hash
     end
   end
 
+  def to_json
+    { date: date_time.to_s,
+
+      total_sp: story_points.total.round,
```
cornelius: Don't use abbreviations in names. This makes the code or the data structures much harder to read and understand. So better call it `total_story_points`.

Same applies to the other names.

cornelius:
The JSON is very similar to the YAML we already have. Do we really need both or could we simplify it by using one for both use cases?

spec/unit/burndown_chart_spec.rb
felixsch: You absolutly right

felixsch:
You're right about the similarity of the json and yaml. I implemented this json structure to match the names used on the server side, that's the only reason why I'm duplicating the code and using this key names.

thutterer:
Joining and picking up the conversation as I'm the one whose project depends on this PR :)

@cornelius:
> The JSON is very similar to the YAML we already have. Do we really need both or could we simplify it by using one for both use cases?

Cornelius, I assume you are talking about the `burndown-data-xx.yaml` files? I was thinking the same thing.

@felixsch:
> I implemented this json structure to match the names used on the server side, that's the only reason why I'm duplicating the code and using this key names.

Well, at the current state of development on my feature, I do not really care yet how the data exactly looks like or what the names of the keys are. I think I could easily process the hole yaml file daily as it is.
I see even an advantage in that approach: If the push to the endpoint failed (for whatever reason) on one day, it could still get that days data the next day.
With this PR's current approach, that's not the case iiuc.

Felix, I guess you implemented it the way you did because @kalabiyau told you to make life (my project) simpler for me. But I really don't care and I think we can avoid the data and code duplications.

Artem, please correct me if I have wrong assumptions.

felixsch:
So I removed to_json. Push to api now sends the hash as json to a host

felixsch:
After speaking with @thutterer and @kalabiyau I changed the data which is being send to endpoint to send all the available data from the yaml file.

thutterer:
@felixsch:
I'm using this code already and it works well for my needs.
Just an idea for improvement... you could allow another argument for `--push-to-api` that specifics the number of a sprint. This way we could use your feature to feed my database with data from older sprints. Afterwards we had the data from all previous sprints in the rails db as-well.

PR #21, opened 2015-11-02
Add make_cover method to set an existing image as cover

thardeck:
It uses the image name to find the correct attachment id.


cornelius:
Rebased and merged

PR #23, opened 2016-02-26
Graph for unplanned cards

aosthof:
- Burndown chart reflects unplanned work
- Draw legend conditionally and fix unplanned graph for fast lane


aosthof:
Unfortunately I wasn't able yet to also create/adjust the necessary tests, this still has to be done.

cornelius:
I added a test and merged a rebased version to master.

PR #26, opened 2016-03-17
Remove unused empty lib/result.rb

promulo:
I think `lib/result.rb` was left behind after some refactoring. :)


cornelius:
Thanks for spotting this and cleaning up. :smile:

PR #27, opened 2016-03-17
Refactor TrelloWrapper class

promulo:
A little bit of refactoging on `TrelloWrapper` class.


spec/unit/trello_wrapper_spec.rb
```diff
@@ -47,6 +47,36 @@
...
   end
 
+  describe '#backup' do
+    it 'raises an error for empty board id' do
+      expect { subject.backup("") }.to raise_error(TrolloloError)
+    end
+
+    it 'raises an error for nil board id' do
+      expect { subject.backup(nil) }.to raise_error(TrolloloError)
+    end
```
cornelius: For consistency there also should be a test for successful execution of the `backup` method as it's done for the `retrieve_board_data` method. Or there should be none.

It's actually debatable if testing that the correct HTTP request is done is useful on this level, because the test mostly just duplicates the implementation and it's unlikely to break by a change unrelated to changing the request itself (which would require a change of the test as well). Maybe it would be better to rely on tests on a higher level which use these methods and then actually process the data returned by these calls.

promulo: Differently from `backup`, `retrieve_board_data` actually does something with the value returned (i.e. convert to a hash), that's why I have included that case. But maybe the correct course of action instead would be to assert that an error is raised because the returned value could not be translated to a hash.

As you said, the granularity here may be too small, indeed.

promulo:
I have removed the positive test case for `retrieve_board_data` method. As @cornelius mentioned, it makes sense to rely on higher level tests for those cases. It also keeps consistency with `backup` tests.

lib/trello_wrapper.rb
```diff
@@ -82,11 +75,24 @@ def set_description(card_id, description)
 
   private
 
+  def client
```
jimmykarily: [minor] I think we can safely memoize this method. No need to create a new object every time ```client``` is called.

promulo: Thanks for the feedback @jimmykarily, and sorry about the long delay. This is a rather old branch, so I have rebased it against current master and memoized the Trello client as you suggested. :slightly_smiling_face: 

Not sure if it's suitable to be merged after this long. Hopefully it still is.

manno:
I think there is something wrong with your rebase, it duplicates commits which are already in the tree?

promulo:
Hi @manno, I believe that happens because GitHub's pull requests are based on the common ancestor of the base branch and the new branch, and since I didn't force push after rebasing, it shows all changes to the base branch in the meantime (it's a pretty old PR :). You'll notice that the commit checksums are the same. Github will just ignore those in case it gets merged.

I know it's annoying but wanted to avoid force-pushing.

manno:
I'd prefer rebase and force push, or a new pull request. This looks like a merge. If I merge this from the command line the resulting tree (git log --graph) looks very confusing to me.

Can I ask you to update this, otherwise I could just cherry pick the commits..

lib/trello_wrapper.rb
```diff
@@ -19,6 +19,7 @@
 class TrelloWrapper < TrelloService
 
   attr_accessor :board
+  attr_accessor :client
```
manno: I think this line is not needed, Ruby will always call the method as it has the same name and is defined later.

promulo: You're right. I will fix. Thanks.

promulo:
Ok @manno, I will then create a new PR with the cherry-picked commits.

promulo:
Hi @manno, there you go #60 :)

promulo:
Closed in favor of #60

PR #28, opened 2016-03-17
Refactor Backup class

promulo:
From my point of view, a `Backup` object should relate to the board it's been backed up. That's why I have included the board id as a parameter in the constructor (and thus as an instance variable). The backup directory was also changed to be a constant.

@cornelius I would appreciate your input here as well :)

Thanks!

### DIFF
--- a/lib/backup.rb
+++ b/lib/backup.rb
@@ -1,35 +1,34 @@
 class Backup
+  BACKUP_DIR = File.expand_path("~/.trollolo/backup")
 
+  attr_reader :board_id
   attr_accessor :directory
 
-  def initialize settings
+  def initialize(board_id, settings)
+    @board_id = board_id
     @settings = settings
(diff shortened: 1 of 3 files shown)


cornelius:
I don't think it's correct to make the board id an attribute of the object. This adds more state than necessary to the class. The class is not meant to handle only one board. It handles backup of any boards. That's why it has a method to list backups.

promulo:
Thanks for the feedback, @cornelius. The reasoning behind that decision was that every time the `Backup` class is instantiated, there is a `board_id` associated. A backup object never handles more than one board at the same time. The exception, of course, is the `list` method, therefore made static.

cornelius:
I guess it's a matter of taste, if you see the class representing the backup of one board or the backups of all board. I actually prefer seeing it as the representation of all backups, so that no exception with a static method is needed, and there is less local state in the class. So let's keep it the way it is.

Issue #24, opened 2016-03-02
Scaling tasks does not always work


cornelius:
Scaling the tasks in a way that the starting point is the same as for story points does not always work. The burndown-23.png and burndown-56.png in the integration tests show that issue. There the green task line should start at the same point as the black story point line.

It probably has something to do with adding tasks during the sprint.

PR #29, opened 2016-03-21
Fix issue #24: Scaling tasks does not always work

aosthof:
In the former implementation, the scalefactor used to get both the story points and tasks graphs starting at the same point was calculated with the maximum values of story points and tasks in a sprint. This led to a changing graph and also a possible offset, when new story points and/or tasks were added during a sprint.

This is now fixed by using the values of total story points and total tasks at the beginning of a sprint (day 0) to calculate the needed scalefactor.


cornelius:
Have you forgotten to push the new versions of the reference images? The integration tests fail here, because the images (23, 31, 42, 56) are not updated to the fixed behavior.

aosthof:
Superseded #31

PR #32, opened 2016-06-20
avoid division by zero

bmwiedemann:


scripts/burndowndata.py
```diff
@@ -155,6 +155,9 @@ def calculateYRange(self, total_story_points, bonus_tasks_done, bonus_story_poin
     return
 
   def setScaleFactor(self, total_tasks, total_story_points):
-    self.scalefactor = float(total_tasks) / float(total_story_points)
+    if total_story_points <= 0.001:
```
Ana06: Good catch, but total_story_points should be and integer, right? So you don't need to compare with 0.001, comparison with 1 should be enough. 

Also, although you didn't wrote it, I guess the `float` method was introduced to ensure that the result is a Float, so I would use `fdiv` instead, which is more readable and elegant

so, in short you could write this like:

``` Ruby
self.scalefactor = total_story_points < 1 ? 1 : total_tasks.fdiv(total_story_points)
```

Ana06:
@bmwiedemann please rebase! :wink:

bmwiedemann:
rebased (was trivial)

Ana06:
@bmwiedemann 

> rebased (was trivial)

yes, we just introduced rubocop, so just to ensure not other offenses are introduced :wink: 

did you see the requested changes? :smile:

scripts/burndowndata.py
bmwiedemann: story_points were intended to be float, e.g. because the trello browser plugin had a "0.5" button
and there can easily be cards smaller than a 1
and re-educating teams to use larger story point values is not a good option and would it hard to compare sprint velocity to previous sprints

Ana06: ok, fine then

Ana06: ok, fine then

PR #34, opened 2016-07-07
sketch out move cards feature

manno:
@cornelius somewhat like this?


cornelius:
Yes. There needs to be an additional argument for the target board, as the cards are moved from the sprint board to the planning board. But other than that the structure looks right.

manno:
replaced by https://github.com/openSUSE/trollolo/pull/35

PR #35, opened 2016-07-18
sprint cleanup command

manno:
Add sprint cleanup command to move remaining cards backs to planning board.


lib/cli.rb
```diff
@@ -322,6 +322,19 @@ def list_member_boards
     }
   end
 
+  desc "after-planning", "Move remaining cards to backlog"
+  long_desc <<EOT
+EOT
```
cornelius: There is a description missing. There either should be one or we don't need the `long_desc` tag, do we?

spec/unit/sprint_cleanup.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  context "default" do
+    subject { described_class.new(dummy_settings) }
+
+    it "moves remaining cards to target board", vcr: "sprint_cleanup" do
+      expect(subject.cleanup("gyp75UHZ", "neUHHzDo")).to be
+    end
+  end
+end
```
cornelius: This file should be named `sprint_cleanup_spec.rb`, shouldn't it?

lib/sprint_cleanup.rb
```diff
@@ -0,0 +1,43 @@
...
+  def sticky?(card)
+    card.labels.any? { |l| l.name == "Sticky" }
+  end
+
+  def move_cards(source_list)
+    source_list.cards.each do |card|
+      next if sticky?(card)
+      puts %(moving card "#{card.name}" to list "#{target_list.name}")
+      card.move_to_board(@target_board, target_list)
+    end
```
cornelius: It would be nice, if it also would remove the people on the cards and the "under the waterline" labels.

Issue #36, opened 2016-08-05
Add command to move cards from planning to sprint board


cornelius:
It would be nice to have a command to do the recurring task of moving cards from the product backlog on the planning board to the sprint backlog on the sprint board. We need to do this after each planning meeting.

Here are the details:

* Add a command `trollolo move-backlog --planning-board-id=ID1 --sprint-board-id=ID2`
* The planning board will contain a list of cards in a column named "Backlog" with two special cards marking the waterline and the seabed.
* The command should move all cards above the seabed card to the column named "Sprint Backlog" on the sprint board, including the waterline card.
* The order of the card including the position of the waterline card has to be preserved.
* All card between the waterline card and the seabed card should get a label "Under waterline"
* The seabed card should be deleted
* If the "Sprint Backlog" column already contains a waterline or seabed card before the move, they should be deleted or moved in a way that the order of the cards in the sprint backlog is correct as it would be if there would not have been these cards.
* If there are any other cards in the "Sprint Backlog" column all moved cards should be appended after them
* A waterline card is a card which has the word "waterline" in the title. The title can be spelled all uppercase, contain additional characters or white space between one or more letters. "Waterline", "water line", and "~ ~ ~ ~ W A T E R L I N E ~ ~ ~ ~" are all valid waterline card titles.
* A seabded card is a card which has the word "seabed" in the title. The same variations are valid as for the waterline card.
* If the "Backlog" column on the planning board does not include both, the waterline and the seabed card, an error should be raised and no card should be moved.
* If the sprint board does not have a column "Sprint Backlog" an error should be raised and no card should be moved.

cornelius:
Done. Thanks, Mario.

PR #37, opened 2016-08-10
Move backlog feature

manno:
Implements Issue #36

This changes rspec vcr integration slightly. The new scrum functionality was moved into a sub folder, as it uses `ruby-trello` directly (most of the time).

Seabed and Waterline card are either moved or if a card exists in the sprint board deleted from the planning board. So styled cards in the sprint backlog should stay the same.
I expect card titles to be uniq and seabed/waterline cards to only exist once in a list.


cornelius:
This looks good. I'll merge it.

The naming of the classes could maybe made a bit more consistent. We have `Prioritizer`. Maybe we should also have `BacklogMover` and `SprintCleaner`. We can change that after the merge, if this makes sense.

PR #39, opened 2016-09-07
Use board aliases

jimmykarily:
With board aliases defined in `trollolorc` file, the users will be able to use "easy" names instead of actual ids for the various commands (for --board-id option).


cornelius:
Nice and clean implementation. Thanks.

PR #41, opened 2016-09-26
Integrate new 'Accepted' column

aosthof:


cornelius:
It would be good to have tests for this code.

It also makes a number of other tests fail, because the test data has no "Accepted" column. We should deal with this more gracefully and not require boards to have an "Accepted" column.

cornelius:
I fixed the merge conflict, added tests, fixed the failing tests, and merged the pull request.

Issue #44, opened 2016-10-26
set-priorities command fails with cards having '#' in the title


cornelius:
If a card has a title which include a `#`, such as `(2) Fix bnc#1004697 - Repository setup inconsistent on kiwi images based on SMT clients` the `--set-priorities` command fails with an error such as

```
/usr/lib64/ruby/2.2.0/uri/rfc3986_parser.rb:66:in `split': bad URI(is not URI?): https://api.trello.com/1/cards/<redacted>/name?key=<redacted>&token=<redacted>&value=P7%3A%20%282%29%20Fix%20bnc#1004697 - Repository setup inconsistent on kiwi images based on SMT clients (URI::InvalidURIError)                                                                                                        
        from /usr/lib64/ruby/2.2.0/uri/rfc3986_parser.rb:72:in `parse'                                                                                            
        from /usr/lib64/ruby/2.2.0/uri/common.rb:226:in `parse'                                                                                                   
        from /usr/lib64/ruby/gems/2.2.0/gems/rest-client-1.8.0/lib/restclient/request.rb:276:in `parse_url'                                                       
        from /usr/lib64/ruby/gems/2.2.0/gems/rest-client-1.8.0/lib/restclient/request.rb:280:in `parse_url_with_auth'                                             
        from /usr/lib64/ruby/gems/2.2.0/gems/rest-client-1.8.0/lib/restclient/request.rb:175:in `execute'                                                         
        from /usr/lib64/ruby/gems/2.2.0/gems/rest-client-1.8.0/lib/restclient/request.rb:41:in `execute'                                                          
        from /usr/lib64/ruby/gems/2.2.0/gems/ruby-trello-1.5.1/lib/trello/net.rb:29:in `execute_core'                                                             
        from /usr/lib64/ruby/gems/2.2.0/gems/ruby-trello-1.5.1/lib/trello/net.rb:18:in `try_execute'                                                              
        from /usr/lib64/ruby/gems/2.2.0/gems/ruby-trello-1.5.1/lib/trello/net.rb:10:in `execute'
        from /usr/lib64/ruby/gems/2.2.0/gems/ruby-trello-1.5.1/lib/trello/client.rb:88:in `invoke_verb'
        from /usr/lib64/ruby/gems/2.2.0/gems/ruby-trello-1.5.1/lib/trello/client.rb:29:in `put'
        from /home/cs/suse/trollolo/lib/trello_wrapper.rb:83:in `set_name'
[...]

PR #46, opened 2016-10-30
Fix issue #44 - set-priorities command fails with cards having '#' in the title

bgeuken:
Trollolo has a subset of methods that change data on a trello board. To
do so trollolo generates URLs which can contain data read from the trello
board.
Since those data might contain characters which are invalid in an URI or have
a certain meaning, we need to escape them.

### DIFF
--- a/lib/trello_wrapper.rb
+++ b/lib/trello_wrapper.rb
@@ -15,6 +15,7 @@
 #  To contact SUSE about this file by physical or electronic mail,
 #  you may find current contact information at www.suse.com
 require 'trello'
+require 'erb'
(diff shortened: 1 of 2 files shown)


bgeuken:
btw, tested manually with the ruby-trello gem. Let me know, if you want a test to ensure that the urls in question get's escaped

jimmykarily:
Good one 👍 . I think it would be good to test this, tell me if you need help with it.

manno:
This has been fixed by using the Trello gems API, instead of building the requests ourselves.

PR #47, opened 2016-10-30
Allow defining checklists that shouldn't be counted as tasks

bgeuken:


lib/empty_column.rb
```diff
@@ -1,5 +1,5 @@
 class EmptyColumn < Column
-  def initialize
-    super({"lists" => [], "cards" => []}, nil)
+  def initialize(settings = nil)
```
jimmykarily: If no settings is passed the default ```nil``` will be used. Next line will call the ```Column``` initializer with nil settings which in turn will call the ```Card``` initializer with nil settings. Then [this line](https://github.com/openSUSE/trollolo/pull/47/files#diff-da53d18413bddf22823980373cf4fff2R82) will break (Haven't tried it though).

I think it will be better if we used the same pattern regarding ```settings``` argument in all method signatures. Either use a default value everywhere or not. I think you can easily add a default value of ```nil``` in all methods by modifying the above line as such:

```
  @settings && @settings.no_task_checklists.include?(checklist.name)
```

lib/scrum_board.rb
```diff
@@ -26,7 +26,7 @@ def done_column
   def accepted_column
     accepted_columns = columns.select{|c| c.name =~ @settings.accepted_column_name_regex }
     if accepted_columns.empty?
-      EmptyColumn.new
+      EmptyColumn.new(@settings)
```
jimmykarily: If you use a default value of ```nil``` as per the previous comment, you can skip this.

lib/checklist.rb
```diff
@@ -6,4 +6,16 @@ def initialize(checklist_data)
...
+    @checklist_data["checkItems"]
+  end
+
+  def done_tasks
+    checklist_items.count { |list_item| list_item["state"] == "complete" }
+  end
+
+  def tasks
+    checklist_items.count
+  end
```
jimmykarily: These methods are being tested though card_spec but maybe it is a good idea to introduce the checklist_spec now that we have some methods to test.

Ana06:
@bgeuken ping! :wink:

Ana06:
@bgeuken you need to rebase. And could you please also add this new feature to the new [CHANGELOG Master section](https://github.com/openSUSE/trollolo/blob/master/CHANGELOG.md#master-unreleased)? :wink:

lib/card.rb
```diff
@@ -76,6 +61,12 @@ def checklists
     end
   end
 
+  def filtered_checklists
+    checklists.reject do |checklist|
+      @settings.no_task_checklists.include?(checklist.name)
```
Ana06: @bgeuken as @jimmykarily said [in his comment](https://github.com/openSUSE/trollolo/pull/47/files#r90401530) this will break if `@settings` is `nil`, which you used as default [here](https://github.com/openSUSE/trollolo/pull/47/files#diff-11fdc3f293c56c46a28349b86bba1764R2)

And I also agree with him on this:

> I think you can easily add a default value of nil in all methods

lib/scrum_board.rb
Ana06: @jimmykarily why? it could be different from `nil` or am I missing something?

bgeuken:
@Ana06 Hey, cut me some slack! I have been waiting for reviews on my PRs for a while;-)

I've updated the PR now, added a note to the changelog and even fixed rubocop:tada:

Though I had to adjust the Metrics/CyclomaticComplexity cop in the .rubocop_todo.yml file.

.rubocop_todo.yml
```diff
@@ -282,7 +282,7 @@ Metrics/ClassLength:
 
 # Offense count: 1
 Metrics/CyclomaticComplexity:
-  Max: 9
+  Max: 10
```
Ana06: You shouldn't modify this manually, but regenerate the todo file. It is easier :wink: Anyway, I think we should disable this cop for the same reason [I disabled `Metrics/AbcSize` and `PerceivedComplexity`](https://github.com/openSUSE/trollolo/pull/91), we want to keep the project as simple as possible for new contributors. So keep this as it is :wink:

lib/empty_column.rb
jimmykarily: The linked line seem to have changed (or never made sense ?).  Maybe I have meant [this line](https://github.com/openSUSE/trollolo/pull/47/files#diff-da53d18413bddf22823980373cf4fff2R66) which resembles my proposal.

lib/scrum_board.rb
jimmykarily: I can't remember why I wrote this but looking at it now, I have to agree with you. Ignore my comment.

jimmykarily:
Sorry for the late answer, I am monitoring the repo and I didn't notice I was mentioned on this one.

Ana06:
@jimmykarily 

> Sorry for the late answer, I am monitoring the repo and I didn't notice I was mentioned on this one.

Thanks to take a look! :wink: 

@bgeuken please rebase, there were some changes in the CHANGELOG. :smiley:

lib/empty_column.rb
bgeuken: I changed that line like you suggested. It's just that I rebased afterwards. So you can't see the old state anymore.

bgeuken:
> Ana06 commented
> And fixing the tests is not needed if you squash the commit and it was never broken. :wink:
I was trying things out and forgot to cleanup afterwards;-)

It should be fine now.

lib/checklist.rb
Ana06: @jimmykarily can you maybe create an issue for this? Then we can finally merge this, and this is not strictly needed for this PR. :wink:

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 * `burndown --plot-to-board` always sets the burndown chart as the cover.
   Fix #114.
 * Allow to create and update an Sprint with a custom number. Fix #78.
+* Allow to define checklists that should not be parsed as task lists. Such lists
```
Ana06: You need to put this in `Master (unreleased)`

Ana06:
and please rebase! :see_no_evil:

bgeuken:
@Ana06 done. Would be great if this could be merged soon:-)

PR #49, opened 2016-11-02
Cleanup generation of ScrumBoard#cards

bgeuken:


lib/scrum_board.rb
```diff
@@ -144,11 +144,6 @@ def id
   end
 
   def cards
-    return @cards if @cards
-    @cards = []
-    columns.each do |column|
-      @cards += column.cards
-    end
-    @cards
+    @cards ||= columns.map { |column| column.cards }.flatten
```
jimmykarily: [MINOR] or: ```@cards ||= columns.map(&:cards).flatten```

bgeuken: Fixed. I've also updated other such cases in a44aac5

PR #52, opened 2016-11-30
Adopt to changes of codeclimate-test-reporter gem

bgeuken:
Version 1.0 of codeclimate test reporter introduced changes that
break configurations for older versions and thus requires some
adoptations.

https://github.com/codeclimate/ruby-test-reporter/blob/master/README.md#upgrading-from-pre-10-versions

====

$ bundle exec rspec
W, [2016-11-23T19:38:05.467313 #5382]  WARN -- :       This usage of the Code Climate Test Reporter is now deprecated. Since version
      1.0, we now require you to run `SimpleCov` in your test/spec helper, and then
      run the provided `codeclimate-test-reporter` binary separately to report your
      results to Code Climate.

      More information here: https://github.com/codeclimate/ruby-test-reporter/blob/master/README.md

### DIFF
--- a/spec/unit/spec_helper.rb
+++ b/spec/unit/spec_helper.rb
@@ -1,5 +1,7 @@
-require "codeclimate-test-reporter"
-CodeClimate::TestReporter.start
+#require "codeclimate-test-reporter"
+#CodeClimate::TestReporter.start
+require "simplecov"
+SimpleCov.start
 require_relative '../../lib/trollolo'
 require 'given_filesystem/spec_helpers'
 require 'webmock/rspec'


jimmykarily:
These changes will work with the versions >= 1.0.0 of the gem but we don't force a version in the gemspec. I tried running the specs with gem version 0.6.0 and these settings and they run ok but I'm not sure if these settings work with versions <= 1.0.0. The [docs](https://docs.codeclimate.com/docs/ruby) implicitly suggest that we force a version. Maybe we should add something like this in the Gemfile:

```
gem 'codeclimate-test-reporter', '~=1.0.0'
```

bgeuken:
I've created issue #54 since I can't tell what the right approach to fix this would be. Thanks for looking at this.

bgeuken:
Wrote some tests for my other PR and stumbled over this agains. Thus updated and reopened. Would be cool, if you guys could merge this soon:-)
 
Added the version restriction as requested. Though the change didn't seem to break running the tests.

bgeuken:
Github doesn't let me reopen this one since I force-pushed => #55

PR #53, opened 2016-11-30
Ignore sticky cards when prioritizing

bgeuken:


jimmykarily:
Now that [this issue](https://github.com/openSUSE/trollolo/issues/54) exists, maybe you should remove the first commit.

manno:
This has been fixed by making `Priotizer` a `TrelloService` and using methods from `CardTypeDetection`.

Issue #57, opened 2016-12-08
Add command to create sprint boards from configuration


manno:
A command like `setup-sprint-boards` should create all necessary boards, lists and labels to work with trollolos SCRUM commands.

The names (Ready, Backlog, QA, Sprint Backlog, Doing, ...) should be configurable via `.trollolorc`. The other scrum commands, like sprint clean up and prioritize, should use these new configuration values instead of command line arguments.

```
sprint_backlog_name: Sprint Backlog
sprint_qa_name: QA
sprint_doing_name: Doing
planning_backlog_name: Backlog
planning_ready_name: Ready for Estimation

```

PR #58, opened 2016-12-08
Fix several issues with SCRUM related commands

manno:
I refactored the Prioritizer to use inherit from TrelloService. We can now use the same card detection methods as in the other SCRUM commands.

* Using the methods from the Trello gem we can avoid #46 
* Add QA list to lists being cleaned up by SprintCleaner
* Ignore cards with sticky label in Prioritizer and SprintCleaner
* Remove command line argument from Prioritizer
* Renamed 'Ready' list to match our board (see #57 for suggestion how to handle names in the future)

### DIFF
--- a/lib/scrum/card_type_detection.rb
+++ b/lib/scrum/card_type_detection.rb
@@ -1,5 +1,9 @@
 module Scrum
   module CardTypeDetection
+    def sticky?(card)
+      card.labels.any? { |l| l.name == "Sticky" }
+    end
+
     def waterline?(card)
       card.name =~ /w.?a.?t.?e.?r.?l.?i.?n.?e/i
     end
--- a/lib/scrum/prioritizer.rb
+++ b/lib/scrum/prioritizer.rb
@@ -9,12 +9,11 @@ module Scrum
     private
 
     def update_priorities
-      priority_name = PriorityName.new
       n = 1
       @board.backlog_cards.each do |card|
         next if @board.sticky?(card) || @board.waterline?(card)
         puts %(set priority to #{n} for "#{card.name}")
-        card.name = priority_name.build(card.name, n)
+        card.name = PriorityName.build(card.name, n)
         card.save
         n += 1
       end
--- a/lib/scrum/prioritizer.rb
+++ b/lib/scrum/prioritizer.rb
@@ -1,36 +1,23 @@
 module Scrum
-  class Prioritizer
-
-    def initialize(settings)
-      @settings = settings
-    end
-
-    def prioritize(board_id, list_name)
-      list = find_list(board_id, list_name)
-      fail "list not found on board" unless list
-      update_priorities(list)
+  class Prioritizer < TrelloService
+    def prioritize(board_id)
+      @board = SprintPlanningBoard.new.setup(board_id)
+      fail "list named 'Backlog' not found on board" unless @board.backlog_list
+      update_priorities
     end
 
     private
 
-    def trello
-      @trello ||= TrelloWrapper.new(@settings)
-    end
-
-    def update_priorities(list)
+    def update_priorities
+      priority_name = PriorityName.new
       n = 1
-      list.cards.each do |card|
-        next if card.name =~ /waterline/i
-        card.priority = n
-        trello.set_name(card.id, card.name)
+      @board.backlog_cards.each do |card|
+        next if @board.sticky?(card) || @board.waterline?(card)
         puts %(set priority to #{n} for "#{card.name}")
+        card.name = priority_name.build(card.name, n)
+        card.save
         n += 1
       end
-    end
-
-    def find_list(board_id, list_name)
-      board = trello.board(board_id)
-      board.columns.find { |list| list.name == list_name }
     end
   end
 end
--- a/lib/scrum/priority_name.rb
+++ b/lib/scrum/priority_name.rb
@@ -2,12 +2,12 @@ module Scrum
   class PriorityName
     PRIORITY_REGEX      = /^(?:\([\d.]+\) )?P(\d+): /
 
-    def priority(name)
+    def self.priority(name)
       return unless m = name.match(PRIORITY_REGEX)
       m.captures.first.to_i
     end
 
-    def build(name, n)
+    def self.build(name, n)
       return name.sub(/P\d+: /, "P#{n}: ") if priority(name)
       "P#{n}: #{name}"
     end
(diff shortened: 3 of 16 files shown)


lib/scrum/priority_name.rb
```diff
@@ -0,0 +1,15 @@
+module Scrum
+  class PriorityName
+    PRIORITY_REGEX      = /^(?:\([\d.]+\) )?P(\d+): /
```
jimmykarily: I think the dot "." shouldn't be there.

The existing regex:
```

2.3.0 :025 > "(.) P3: ".match(/^(?:\([\d.]+\) )?P(\d+): /)
 => #<MatchData "(.) P3: " 1:"3"> 
```

without the dot:

```
2.3.0 :027 > "(.) P3: ".match(/^(?:\(\d+\) )?P(\d+): /)
 => nil 
2.3.0 :028 > "(3) P3: ".match(/^(?:\(\d+\) )?P(\d+): /)
 => #<MatchData "(3) P3: " 1:"3">
```

I don't think we want to catch this case: "(.) P3:"

lib/scrum/priority_name.rb
```diff
@@ -0,0 +1,15 @@
+module Scrum
+  class PriorityName
```
jimmykarily: This class has no initializing code, no instance variable and no state. Only two helper methods. Maybe it is better to just make them class methods. This way we won't have to initialize this class before using it (for no reason).

lib/scrum/priority_name.rb
```diff
@@ -0,0 +1,15 @@
...
+    PRIORITY_REGEX      = /^(?:\([\d.]+\) )?P(\d+): /
+
+    def priority(name)
+      return unless m = name.match(PRIORITY_REGEX)
+      m.captures.first.to_i
+    end
+
+    def build(name, n)
```
jimmykarily: This method implies that there should be no card which doesn't have priority set but has an estimation. To explain:

- Lets say I have this card "(3) Prepare dinner for my beloved one"
- I pass it through ```build``` method here:

```
  card.name = priority_name.build(name, 8)
```
- Now ```card.name``` is "P8: (3) Prepare dinner for my beloved one" which unfortunately doesn't match the priority regex (although ```build``` was supposed to return a "valid" name).

We can either create a convention that we don't support card with estimation but no priority (but I'm not sure if it fits all the flows) or change this method a bit so it takes care of that case too.

Whichever the case, I think a test should be added, that will check that either an error is returned when a card with an invalid name is detected or that the case is handled gracefully.

lib/scrum/sprint_cleaner.rb
```diff
@@ -1,15 +1,16 @@
 module Scrum
   class SprintCleaner < TrelloService
-    SOURCE_LISTS = ["Sprint Backlog", "Doing"]
-    TARGET_LIST = "Ready"
+    TARGET_LIST = "Ready for Estimation"
```
jimmykarily: This means that anyone using trollolo should change the name of this list right? We should at least add a record in the CHANGELOG to warn everybody.

lib/scrum/priority_name.rb
jimmykarily: I now saw that there is a test for this. Actually the "P8: (3)" format is also valid. Ignore this one.

spec/unit/scrum/priority_name_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it "updates priority after story points" do
+      expect(
+        priority_name.build("(0.5) P1: Refactor cards", 4)
+      ).to eq("(0.5) P4: Refactor cards")
+    end
+
+    it "adds priority after story points" do
+      expect(
+        priority_name.build("(0.5) P1: Refactor cards", 5)
+      ).to eq("(0.5) P5: Refactor cards")
```
jimmykarily: This is the same as the previous one (although the description says otherwise). I don't think it ever adds priority after story points.

lib/scrum/prioritizer.rb
```diff
@@ -1,36 +1,23 @@
...
-      list.cards.each do |card|
-        next if card.name =~ /waterline/i
-        card.priority = n
-        trello.set_name(card.id, card.name)
+      @board.backlog_cards.each do |card|
+        next if @board.sticky?(card) || @board.waterline?(card)
         puts %(set priority to #{n} for "#{card.name}")
+        card.name = priority_name.build(card.name, n)
+        card.save
         n += 1
```
manno: I prefer to have the right number clearly visible in `n`. I'd save two lines, but introduce two additions by relying on the zero based index.

lib/scrum/card_type_detection.rb
```diff
@@ -1,5 +1,9 @@
 module Scrum
   module CardTypeDetection
+    def sticky?(card)
+      card.labels.any? { |l| l.name == "Sticky" }
```
manno: Would be great if I'd  be able to use it, but `include?` won't look at the `name` property.

lib/scrum/sprint_board.rb
```diff
@@ -16,6 +16,22 @@ def backlog_list_name
...
+    def doing_list_name
+      "Doing"
+    end
+
+    def doing_list
+      @board.lists.find { |l| l.name == doing_list_name }
+    end
+
+    def qa_list
+      @board.lists.find { |l| l.name == qa_list_name }
```
manno: This is easily done, but ultimately this should be improved upstream.  Well, let's just cache it for now ;)

lib/scrum/priority_name.rb
manno: I guess it shouldn't match dot, but maybe someone is using fractions, like `(1.5) P2:`.
So, an improvement might be `/^(?:\(\d+(?:\.\d+)?\) )?P(\d+): /`?

Anyways, I just extracted these methods and don't plan to improve them with this PR. Additionally `priority` should be a private method, there are no consumers outside `PriorityName`, but since there were already some tests for it I left it public.

lib/scrum/priority_name.rb
manno: Yes, that's a good place to class methods. Honestly I just moved these methods here quickly, because they were part of the enormous `Card` class. 
I thought this might become a real object later, but that's currently not necessary at all.

lib/scrum/sprint_cleaner.rb
manno: I don't plan on creating a release just yet. Also  https://github.com/openSUSE/trollolo/pull/59 will change how names are handled.

manno:
Merged in https://github.com/openSUSE/trollolo/commit/7858f1d2f7176486f32a51a1c68409aa33cbe735

PR #73, opened 2017-05-03
Fix name of the backlog column on Planning Board

aosthof:
The name of the backlog column on our Planning Board has changed from "Backlog" to "Platform Backlog", but that change was not reflected in trollolo yet, therefore moving cards from the planning board to the sprint board didn't work anymore. This commit fixes this issue.

### DIFF
--- a/lib/settings.rb
+++ b/lib/settings.rb
@@ -75,7 +75,7 @@ class Settings
         "sprint_backlog" => "Sprint Backlog",
         "sprint_qa" => "QA",
         "sprint_doing" => "Doing",
-        "planning_backlog" => "Backlog",
+        "planning_backlog" => "Platform Backlog",
         "planning_ready" => "Ready for Estimation"
       },
     }.freeze


manno:
Instead of changing the default configuration you could edit your `.trollolorc`. The available options including `board_names.planning` are described in the [README](https://github.com/openSUSE/trollolo/blob/master/README.md#default-configuration)

cornelius:
Or introduce a command line option to specify the name as we already have it for the set-priorities command.

manno:
I see you introduced that in eaaf5179d6b213c46cea82ca9911ece748e3d1f9 does that also use the `scrum.list_names.planning_backlog` from the config?

Ana06:
@manno 
> I see you introduced that in eaaf517 does that also use the scrum.list_names.planning_backlog from the config manifest?

It is just taking the name specified as parameter, or what do you exactly mean?

Ana06:
I created an issue: https://github.com/openSUSE/trollolo/issues/82

So we can close this

PR #92, opened 2017-09-11
Prepare version 0.0.11

### LINKED ISSUE #72: Show new title of card in output of set-priorities
When running `trollolo set-priorities` the cards being prioritized are listed on stdout. It should list the new title of the card, not the old one. Then the new priorities already show up in the titles.

This can even work as a poor man's version of #45 then.

### LINKED ISSUE #77: Being able to specify the sprint length and weekend_lines
If I am not wrong currently there is no way to specify in the command line the `total_days` and `weekend_lines`. So that means that if I want to have an sprint with 8 days I have to generate the .yml, then manually modify the .yml and after that painting the graph again. This could be easily done with two more parameters for `trollolo burndown --new-sprint`. This would also help to automatize the generation of the burndown-char.

Ana06:
- Add in `trollolo burndown --new-sprint` comand the `total_days` and `weekend_lines` params. Closes #77.
- Change stdout output when running set-priorities to render the new priority instead of the old one. Fixes #72.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Trollolo Changelog
 
+## Version 0.0.11
+
+* Add in `trollolo burndown --new-sprint` command the `total_days` and
+  `weekend_lines` params. Closes #77.
+* Change stdout output when running set-priorities to render the new priority
+  instead of the old one. Fixes #72.
+
 ## Version 0.0.10
 
 * Rename `sprint-cleanup` to `cleanup-sprint`.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Trollolo Changelog
 
+## Version 0.0.11
+
+* Add in `trollolo burndown --new-sprint` comand the `total_days` and
```
evanrolfe: [Nitpick] typo: comand => command

Issue #96, opened 2017-09-13
Remove Ruby 2.1 support


Ana06:
I built and installed `trollolo` with Ruby 2.1 and it is not working:

``` console
Ana@linux-nt5w:~/Documents/trollolo(master)> gem.ruby2.1 build trollolo.gemspec 
WARNING:  pessimistic dependency on ruby-trello (~> 1.5.0) may be overly strict
  if ruby-trello is semantically versioned, use:
    add_runtime_dependency 'ruby-trello', '~> 1.5', '>= 1.5.0'
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: trollolo
  Version: 0.0.11
  File: trollolo-0.0.11.gem

Ana@linux-nt5w:~/Documents/agile-team-dashboard(burndown-chart)> vagrant exec sudo gem.ruby2.1 install trollolo
Building native extensions.  This could take a while...
ERROR:  Error installing trollolo:
        ERROR: Failed to build gem native extension.

    current directory: /usr/lib64/ruby/gems/2.1.0/gems/unf_ext-0.0.7.4/ext/unf_ext
/usr/bin/ruby.ruby2.1 -r ./siteconf20170913-4441-1e8v44t.rb extconf.rb
mkmf.rb can't find header files for ruby at /usr/lib64/ruby/include/ruby.h

extconf failed, exit code 1

Gem files will remain installed in /usr/lib64/ruby/gems/2.1.0/gems/unf_ext-0.0.7.4 for inspection.
Results logged to /usr/lib64/ruby/gems/2.1.0/extensions/x86_64-linux/2.1.0/unf_ext-0.0.7.4/gem_make.out
```

I would suggest to remove Ruby 2.1 support as it is quite old. This way we can also remove [activesupport dependency](https://github.com/openSUSE/trollolo/blob/master/trollolo.gemspec#L21) which seems to be introduced only for staying compatible with Ruby 2.1: https://github.com/openSUSE/trollolo/commit/1b97fc55fcddeea9d7700121cff93496ba66a395

PR #97, opened 2017-09-13
Remove Ruby 2.1 support

Ana06:
It is not working and this allow us to remove the activesupport  dependency.

Closes https://github.com/openSUSE/trollolo/issues/96.


Ana06:

Issue #83, opened 2017-09-08
Document all trollolo option


Ana06:
All the commands that Trollolo has, are not documented here: https://github.com/openSUSE/trollolo/blob/master/man/trollolo.1.md#commands

**Clue:** The Trollolo available commands can be get running `trollolo help`

alexandergraul:
Working on it.

PR #99, opened 2017-09-13
Add documentation for CLI commands

alexandergraul:
Add documentation for all available CLI commands to the manpage.

Fix https://github.com/openSUSE/trollolo/issues/83.

### DIFF
--- a/man/trollolo.1.md
+++ b/man/trollolo.1.md
@@ -36,6 +36,13 @@ and cards and has functionality for extracting data for burndown charts.
 
 ## COMMANDS
 
+### help -- Display help
+
+`trollolo help [COMMAND]`
+
+Displays help about the available trollolo commands. Can take a command as 
+an argument to display detailed information about it.
+
 ### burndown-init -- Initialize burndown chart
 
 `trollolo burndown-init --board-id=<board id> --output=<directory>`
@@ -66,6 +73,13 @@ For correct generation of the burndown chart, the Trello board has to follow
 a few convention. They are described in the section `CONVENTIONS for SCRUM
 BOARDS`.
 
+### burndowns -- Process data for multiple burndown charts at once
+
+`trollolo burndowns --board-list=<board list>`
+
+Updates the burndown data for all boards specified in the YAML file in the 
+given directory. See the previous section for details on the update process. 
+
 ### plot -- Plot burndown chart
 
 `trollolo plot <sprint-number>`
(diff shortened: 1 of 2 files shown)


man/trollolo.1.md
```diff
@@ -36,6 +36,13 @@ and cards and has functionality for extracting data for burndown charts.
 
 ## COMMANDS
 
+### help -- Display help
+
+`trollolo help [COMMAND]`
+
+Display help and exit. Can take a command as an argument to display detailed
```
Ana06: I don't think it is needed saying that the command exit :wink: and it should be display**s** . What aboud this:

_Displays help about the available trollolo commands. Can take a command as an argument to display detailed information about it._

man/trollolo.1.md
```diff
@@ -83,16 +97,124 @@ conventions for Scrum boards, and reports burndown data.
...
+Create boards, lists and labels with names configured in '~/.trollolorc' or
+with the defaults.
+
+### move-backlog -- Move backlog from planning to sprint board
+
+`trollolo move-backlog --planning-board-id=<planning board id>
+--sprint-board-id=<sprint board id>`
+
+Move cards from the 'Backlog' list on the planning board to the 'Sprint
+Backlog' list on the sprint board.
```
Ana06: What about mentioning here that this is what is doing at the beginning of the sprint once the planning meeting is finished? :thinking:

man/trollolo.1.md
```diff
@@ -83,16 +97,124 @@ conventions for Scrum boards, and reports burndown data.
...
+
+### cleanup-sprint -- Move cards back to planning board
+
+`trollolo cleanup-sprint --board-id=<board id> --target-board-id=<target board
+id>`
+
+Move cards from 'Sprint Backlog', 'Doing' and 'QA' lists on the sprint board
+to the 'Ready' list on the planning board.
```
Ana06: What about adding that this is what is done when the Sprint finished, after the review, as the unfinished card are moved to the backlog to work on them in future sprints?

man/trollolo.1.md
alexandergraul: I think anyone that uses this tool to manage their Scrum should know when to move the cards from the planning board to the sprint board / the other way around but I can add a sentence for each (`move-backlog` and `cleanup-sprint`)

Ana06: yes, but it is difficult to read. If we add that it is used in this case it is much clear.

Issue #68, opened 2017-02-26
Run burndown generation as part of sprint cleanup


cornelius:
The command `trollolo cleanup-sprint` moves cards back from the sprint board to the planning board. This marks the end of the sprint and prepares for the next one. To make sure that the burndown data is complete and reflects the latests state the cleanup command could run the burndown generation before moving cards.

PR #104, opened 2017-09-14
Run burndown generation on `cleanup-sprint`

alexandergraul:
Add automatic generation of burndown data when calling
`trollolo cleanup-sprint`. This assumes the burndown-data-*.yaml is in
the current working directory. The user gets notified if new data was
generated or not.

Fix https://github.com/openSUSE/trollolo/issues/68.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 * Add a new `plot-to-board` option to the burndown command to send the plotted
   burndown chart to the first card of the `Done` column.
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Trollolo Changelog
 
 ## Master (unreleased)
+* Run `burndown` on `cleanup-sprint`. Fixes #68.
```
Ana06: Just for coherence, leave an empty line before :wink:

PR #105, opened 2017-09-14
Add plot-to-board option to burndown

Ana06:
Add a new option to the burndown command to send the plotted burndown chart to the first card of the `Done` column. :bowtie: 

Pair programmed with @alexandergraul.


Ana06:

PR #113, opened 2017-09-19
Fix burndown --plot-to-board -o

Ana06:
When running `trollolo burndown` with `--plot-to-board` and `-o` options the command failed as it didn't look for the image to upload in the correct directory. :bowtie:


Ana06:

Issue #114, opened 2017-09-19
Enforce that --plot-to-board make the burndown chart cover


Ana06:
Currently the image is upload to the card as an attachment and some times is added as a cover, but not always. It would be great to enforce that it is always set as a cover.

PR #116, opened 2017-09-22
Enforce that plot-to-board sets chart as cover

alexandergraul:
`trollolo burndown --plot-to-board` only  added the new chart to
the card as an attachment. Trello does not always set the new 
attachment as the cover. This ensures that the attachment is added as a cover.
Fix https://github.com/openSUSE/trollolo/issues/114.

I could not test this code: The tests are broken (failing with current release) 
and I also could not find a test for `trollolo burndown --plot-to-board` to extend.
The change in code is very simple though, just calling one additional "Trello" method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Fix `plot-to-board` option in the burndown command when it is used together
   with `-o`.
+* `burndown --plot-to-board` always sets the burndown chart as the cover.
+  Fix #114.
 
 ## Version 0.0.14
 
--- a/lib/burndown_chart.rb
+++ b/lib/burndown_chart.rb
@@ -215,7 +215,9 @@ class BurndownChart
       board = trello.board(board_id)
       name = options['output'] ? options['output'] : '.'
       name += "/burndown-#{sprint.to_s.rjust(2, '0')}.png"
-      trello.add_attachment(board.burndown_card_id, name)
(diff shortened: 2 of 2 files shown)


Ana06:
@alexandergraul 
> The tests are broken (failing with current release)

what do you mean with this? :confused:

alexandergraul:
@Ana06 I am talking about running the current test for the burndown_chart_spec.rb locally:
```
$ bundle exec rspec spec/unit/burndown_chart_spec.rb
Finished in 0.06937 seconds (files took 0.45787 seconds to load)
26 examples, 14 failures
```
I just looked at Travis, the same tests are passing there. Something must be wrong with my test environment/setup.

Anyway, there is no test for the `--plot-to-board` option, that I could add my changes to. I also don't know how to check my change specifically since trello only sometimes doesn't set the uploaded picture as cover. So just checking if the uploaded picture is not enough (could just be luck that the `trello.add_attachment()` did its job during the test, so my change does not get tested in that case.
Also, `trollolo make-cover` has its own test that covers `trello.make_cover()`, so that part should be working fine.

Should I try to add a test for the `--plot-to-board` option in this PR or rather do it as a standalone PR (since it is closer related to #105 than this change)? I prefer a separate PR, in my opinion that is a different issue (adding test to already implemented code).

Ana06:

Ana06:

Issue #78, opened 2017-09-05
custom number for sprint


Ana06:
This is the documentation for creating a new sprint:

`n, [--new-sprint=NEW_SPRINT]                   # Create new sprint`

Because of that I had though you can have a custom name in the sprint, but that is not the case as the value it is just ignored here: https://github.com/openSUSE/trollolo/blob/master/lib/cli.rb#L179

It should be possible to provide a custom number number for creating a sprint:

`trollolo burndown --new-sprint=7`

Clues:
- It should only allow numbers as parameters.
- It should work and be tested for limit cases, for example: The number is the same of the current sprint.
- It should be documented.

Issue #76, opened 2017-09-05
previous sprint .yml shouldn't be needed


Ana06:
When running `trollolo burndown --new-sprint` the `.yml` from the previous sprint is needed, but we are actually only using the number and the `board_id`. This two parameters can be also added in the command: `trollolo burndown --new-sprint --board-id=Fs7boVwI --new-sprint=3` so in this case the previous sprint shouldn't be needed.

```
trollolo burndown --new-sprint --board-id=Fs7boVwI --new-sprint=3
'/vagrant/burndown-data-01.yaml' not found
```

Ana06:
> And having this solved I would say that running trollolo burndown-init is also not longer needed, right?

This may be true only with https://github.com/openSUSE/trollolo/issues/77

Ana06:
Just for clarification, it shouldn't be needed but should work with it too.

Ana06:
I think that should be included in release 1.0.0

PR #117, opened 2017-10-06
Allow to create an Sprint with a custom number

Ana06:
It should be possible to provide a custom number number for creating a
sprint. As we are currently always updating the sprint with higher
number that also means that we need to be able to modify the sprint for
a concrete Sprint. So, I had implemented a new option `sprint-number`
which can be used for both cases.

Closes https://github.com/openSUSE/trollolo/issues/78

This also contributes to https://github.com/openSUSE/trollolo/issues/76


Ana06:

PR #120, opened 2017-10-06
Prepare version 0.1.0

### LINKED ISSUE #78: custom number for sprint

### LINKED ISSUE #114: Enforce that --plot-to-board make the burndown chart cover

Ana06:
Changes from 0.0.14:

* Fix `plot-to-board` option in the burndown command when it is used together
  with `-o`.
* `burndown --plot-to-board` always sets the burndown chart as the cover.
  Fix #114.
* Allow to create and update an Sprint with a custom number. Fix #78.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master (unreleased)
 
+
+## Version 0.1.0
+
 * Fix `plot-to-board` option in the burndown command when it is used together
   with `-o`.
 * `burndown --plot-to-board` always sets the burndown chart as the cover.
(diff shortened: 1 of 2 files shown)


Ana06:

PR #121, opened 2017-10-06
Fix setting char as cover in plot-to-board

Ana06:
This improvement was introduced in https://github.com/openSUSE/trollolo/pull/116, but it was only working on the case that the file was in the current directory. When the route is different it is not working as the name in the card doesn't include the path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master (unreleased)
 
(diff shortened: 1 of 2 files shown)


Ana06:
@alexandergraul you may want to take a look :wink:

DavidKang (approved):

Ana06:

agraul (approved):
Had troubles understanding the issue at first, but now I got it. Looks good!

PR #122, opened 2017-10-06
Prepare version 0.1.1

Ana06:
* Fix the bug introduced whith always setting the burndown chart as the cover
  for `burndown --plot-to-board`, as it only worked with files in the current
  directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Master (unreleased)
 
+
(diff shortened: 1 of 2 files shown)


DavidKang (approved):

Ana06:
[(3) Frontpage: Burndown Chart](https://trello.com/c/Gm0FGST1/539-3-frontpage-burndown-chart)

Issue #100, opened 2017-09-14
board-id in burndown is ignored


Ana06:
When running `trollolo burndown --new_sprint --board-id=78454 --output=trollolo` the `board-id` parameter is ignored and it uses the `board-id` of the previous sprint, which is in the last `burndown-data.yml` file.

alexandergraul:
I'll work on this.

alexandergraul:
for clarification: isn't this a new feature?
right now there is no `--board-id` for `trollolo burndown`.

Ana06:
I think it appeared, at least some weeks ago, when running `trollolo --help`. Anyway, it is not important, if it is not implemented at all, we could implement it, it is already in other commands. :wink:

Ana06:
@alexandergraul how are you going with this? :thinking:

alexandergraul:
@Ana06 I will work on this in the upcoming days, I assume I will have a PR ready in the first half of the next week.

PR #125, opened 2017-10-09
add --board-id to trollolo burndown

agraul:
Add `trollolo burndown --board-id` to override previous `board id`
in the `burndown-*.yaml`. The given `board id` is saved in the updated
`burndown-*.yaml`.

I had to run `rubocop --auto-gen-config` to exclude `Metrics/MethodLength` and `Metrics/BlockLength`
after adding tests and webmock data. I made sure that rubocop had nothing else to complain about first.

Fix https://github.com/openSUSE/trollolo/issues/100.

### DIFF
--- a/lib/burndown_chart.rb
+++ b/lib/burndown_chart.rb
@@ -195,6 +195,7 @@ class BurndownChart
   def update(options)
     burndown_data_path = load_sprint(options['output'] || Dir.pwd, options[:sprint_number])
 
+    @data['meta']['board_id'] = options['board-id'] if options.has_key?('board-id')
     burndown_data = BurndownData.new(@settings)
     burndown_data.board_id = board_id
     burndown_data.fetch
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -175,6 +175,7 @@ EOT
   option 'with-fast-lane', desc: 'Plot Fast Lane with new cards bars', required: false, type: :boolean
   option 'no-tasks', desc: 'Do not plot tasks line', required: false, type: :boolean
   option 'push-to-api', desc: 'Push collected data to api endpoint (in json)', required: false
+  option 'board-id', desc: 'Id of Trello Board'
   def burndown
     process_global_options options
     require_trello_credentials
--- a/spec/unit/burndown_chart_spec.rb
+++ b/spec/unit/burndown_chart_spec.rb
@@ -408,7 +408,7 @@ EOT
 
     describe 'update' do
       let(:path) { given_directory_from_data('burndown_dir') }
-      let(:options) { {'output' => path} }
+      let(:options) { {'output' => path, 'board-id' => '7Zar7bNm'} }
       let(:before) { BurndownChart.new(@settings) }
       let(:after) { BurndownChart.new(@settings) }
 
@@ -432,6 +432,11 @@ EOT
         @chart.update(options)
         after.read_data(File.join(path, 'burndown-data-02.yaml'))
         expect(after.days.size).to eq before.days.size + 1
+      end
+
+      it 'uses provided board-id' do
+        @chart.update(options)
+        expect(@chart.board_id).to eq '7Zar7bNm'
       end
     end
 
--- a/spec/unit/support/webmocks.rb
+++ b/spec/unit/support/webmocks.rb
@@ -28,6 +28,15 @@ def webmock_mapping
         'card_checklists' => 'all'
       },
       file: 'full-board-with-accepted.json'
+    },
+    {
+      path: 'boards/7Zar7bNm',
+      parameters: {
+        'cards' => 'open',
+        'lists' => 'open',
+        'card_checklists' => 'all'
+      },
+      file: 'full-board.json'
     }
   ]
 end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,6 +21,8 @@ Metrics/LineLength:
 # TODO: Down it to 30, 3 times more than Rubocop default
 Metrics/MethodLength:
   Max: 35
+  Exclude:
+    - 'spec/unit/support/webmocks.rb'
 
 Metrics/PerceivedComplexity:
   Enabled: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   is optional and defaults to the current working directory. Fix #103.
 * Allow to define checklists that should not be parsed as task lists. Such lists
   can be added in the trollolorc as `no_task_checklists`.
+* Allow to provide a board id when calling `burndown`. Fix #100.
 
 ## Version 0.1.1
 


lib/burndown_chart.rb
```diff
@@ -195,6 +195,10 @@ def load_sprint(burndown_dir, number = nil)
   def update(options)
     burndown_data_path = load_sprint(options['output'] || Dir.pwd, options[:sprint_number])
 
+    if options.has_key?('board-id')
```
Ana06: why adding this in the hash? You can directly do this in the assignment. Please move this to https://github.com/openSUSE/trollolo/pull/125/files#diff-17ddcbd394dd49848e146577f8d3d845L199

It would be like:

``` Ruby
burndown_data.board_id = options['board-id'] || board_id

File: lib/cli.rb [hunk 2]
```diff
@@ -175,6 +175,7 @@ def burndown_init command = nil
   option 'with-fast-lane', :desc => "Plot Fast Lane with new cards bars", :required => false, :type => :boolean
   option 'no-tasks', :desc => "Do not plot tasks line", :required => false, :type => :boolean
   option 'push-to-api', :desc => 'Push collected data to api endpoint (in json)', :required => false
+  option 'board-id', :desc => "Id of Trello board", :required => false
```
Ana06: `:required => false` is the default, please remove it

and use new Ruby syntax and single quotes if there is not string interpolation:

``` Ruby
option 'board-id', desc: 'Id of Trello board'

File: spec/unit/burndown_chart_spec.rb [hunk 3]
```diff
@@ -407,8 +407,9 @@
     end
 
     describe "update" do
+
       let(:path) { given_directory_from_data("burndown_dir") }
-      let(:options) { {'output' => path} }
+      let(:options) { {'output' => path, 'board-id' => "7Zar7bNm"} }
```
Ana06: use single quotes! :sob:

spec/unit/burndown_chart_spec.rb
```diff
@@ -433,6 +434,11 @@
         after.read_data(File.join(path, 'burndown-data-02.yaml'))
         expect(after.days.size).to eq before.days.size + 1
       end
+
+      it "uses provided board-id" do
+        @chart.update(options)
+        expect(@chart.board_id).to eq "7Zar7bNm"
```
Ana06: single quotes :pray:

spec/unit/support/webmocks.rb
```diff
@@ -28,6 +28,15 @@ def webmock_mapping
         "card_checklists" => "all"
       },
       file: 'full-board-with-accepted.json'
+    },
+    {
+      path: 'boards/7Zar7bNm',
+      parameters: {
```
Ana06: singles quotes everywhere here!

Ana06 (changes requested):
@alexandergraul thanks for your PR! :smile: 

> I had to run rubocop --auto-gen-config to exclude Metrics/MethodLength and Metrics/BlockLength
after adding tests and webmock data. I made sure that rubocop had nothing else to complain about first.

It makes sense to disable `Metrics/BlockLength` for rspec. I already sent a PR for this: https://github.com/openSUSE/trollolo/pull/126. But sorry, I do not agree with `Metrics/MethodLength`, a method with more than 30 lines is for sure not readable. So do not regenerate the `todo` file, instead solve the offense. :wink:

Ana06:
@alexandergraul https://github.com/openSUSE/trollolo/pull/126 got merged, so you don't need to worry about `Metrics/BlockLength` in rspec anymore. Please remove your changes in the `.rubocop_todo.yml` file. :wink:

alexandergraul:
Hi, thanks for your comments. For the quotes I thought it would be better to stick to what is currently the "norm" in the code, but I can also change it to single quotes. In my opinion it looks better unified than suddenly enforcing one style though I have no objection to changing it.

I do however think the 40 line "method" should stay unless `webmocks.rb` gets excluded from that cop. `def webmock_mapping` had 35 lines previously. I can't see how my 5 lines of code (which are needed to mock the test for my test board - until there are major changes to the test suite I see no other way to test this) made that method less readable at all. I don't know how to split that into two methods either, but I think that a method length enforcement for that method is wrong, so I did the 'quick fix'. I would not generally create a new config to avoid offenses, but at the current state of the test suite I saw no other way. Actually, thinking about it now I probably could have excluded that file from that cop instead of changing the max lines, but even then the limit would be at 35 which you think (and I agree 100%) is too long. I guess me changing this gives this wrongly configured cop exposure, so that is good :)

I'll integrate the other changes tomorrow :)

lib/burndown_chart.rb
alexandergraul: I did not remember at first, but now it I do: that does not work. I tried it like that first, but it does not update the board-id in the burndown.yaml which lead me to the solution you can see above (which does work)

lib/cli.rb
alexandergraul: I disagree with the suddenly using single quotes when the existing codebase uses a different style. Except for `push-to-api`, every option uses double quotes for the description.

lib/burndown_chart.rb
Ana06: ok... :see_no_evil: 

Then write it at least in one single line following [Ruby Style guide](https://github.com/bbatsov/ruby-style-guide#if-as-a-modifier), so:

``` Ruby
@data['meta']['board_id'] = options['board-id'] if options.has_key?('board-id')
```

lib/cli.rb
Ana06: We should change all of them. If you want you can include this in a different commit in this PR. But if not, we already have a Rubocop cop that will be enabled. But please do not introduce more offences.

.rubocop_todo.yml
```diff
@@ -1,12 +1,12 @@
 # This configuration was generated by
 # `rubocop --auto-gen-config`
-# on 2017-09-14 15:05:31 +0200 using RuboCop version 0.49.1.
```
Ana06: Do not regenerate this file, you shouldn't include new offences!!! :angry:

Ana06 (changes requested):
@alexandergraul 

> Hi, thanks for your comments. For the quotes I thought it would be better to stick to what is currently the "norm" in the code, but I can also change it to single quotes. In my opinion it looks better unified than suddenly enforcing one style though I have no objection to changing it.

The norm is what he have in our Rubocop config file. That is our style rules. We didn't have time to solve all the offences yet, but this cop support autocorrect, so we can easily enable it. But please do not introduce more offences.

You can just add a new commit enabling the rubocop cop and autocorrect everything. So it is easy to do.

What the problem with splitting `webmocks.rb`?

alexandergraul:
okay, I am still not convinced about changing the style of new code before updating existing code to our new style guide, but this discussion does not lead anywhere.

The problem with `webmocks.rb` is that `webmock_mapping` returns a list of mocked trello boards. Splitting this list into more than one piece only reduces readability and does not improve it. The contents of the list are all very similar and I think it makes sense to group them into one list. It also adds unnecessary complexity to `full_board_mock` which iterates over the mentioned list.

In conclusion, limiting the list to ~30/~35 lines removes clarity from the list ("why is it split at an arbitrary position") and also adds complexity when accessing it.

Ana06:
@alexandergraul 

> okay, I am still not convinced about changing the style of new code before updating existing code to our new style guide, but this discussion does not lead anywhere.

Our code style is defined in Rubocop. It is just that the offenses are not yet totally corrected. But I just solved all the offenses (https://github.com/openSUSE/trollolo/pull/127), it is much faster than discussing about something that is not that important. So please rebase and follow the style. Having an style guide and a style checker (Rubocop) has the objetive to avoid discussing about style in PRs. :wink: 

>The problem with webmocks.rb is that webmock_mapping returns a list of mocked trello boards. Splitting this list into more than one piece only reduces readability and does not improve it. The contents of the list are all very similar and I think it makes sense to group them into one list. It also adds unnecessary complexity to full_board_mock which iterates over the mentioned list.
>
>In conclusion, limiting the list to ~30/~35 lines removes clarity from the list ("why is it split at an arbitrary position") and also adds complexity when accessing it.

I don't really agree, but I understand your point of view. I gave up, you can exclude it. But you should exclude it manually in the `.rubocop.yml` file, you shouldn't regenerate the todo file, because that would means that we want to enable this in the future. So, the `.rubocop_todo.yml` is used when adding Rubocop to big projects for the first time to avoid having to enable all the rules at the same time and knowing what is missed. Our specific configuration, so the things we decide that we want to disable should be in the `.rubocop.yml` file.


And please rebase!

alexandergraul:
@Ana06 Please take a look again, I think I changed everything to your requests.
Now that the rest of the code uses single-quotes, there is no question which one to use anymore  :smile:

TheRealBro:
@alexandergraul 
Don't forget to change README.md please 😃
if you add new "--options" to a software

alexandergraul:
@TheRealBro thanks, I forgot that. added it just now :)

README.md
```diff
@@ -107,6 +107,11 @@ This will get the current data from the Trello board and update the data file
 with the data from the current day. If there already was some data in the file
 for the same day it will be overridden.
 
+To override the Trello board from the `burndown-data-SPRINT_NUMBER.yaml` with another board
```
Ana06: @TheRealBro 

> @alexandergraul 
Don't forget to change README.md please 😃
if you add new "--options" to a software

I don't think this is needed in the README. In the README we have the basic usage, all the option are documented with detail in the [man page](https://github.com/openSUSE/trollolo/blob/master/man/trollolo.1.md).  And the `--board-id` option is already documented there. It is true that the documentation can be improved, but I don't want to overload the README.

@TheRealBro can you open an issue to improve the documentation of the project so we can discuss this there instead of doing it in @alexandergraul PR? :wink: 

In the meanwhile, @alexandergraul can you remove this please? I think it is not needed, and we are not documenting something like that here for any other command, so for consistency as well. :smile:

Ana06 (changes requested):
LGTM apart from removing the documentation from the readme :wink:

Ana06 (approved):
Thanks for the contribution @alexandergraul :heart_eyes:

.rubocop_todo.yml
mus-fever: This is already correct - line 22 uses fstd.organizer_fiscal_code which comes from fiscal_seal_transaction_detail_view. That view uses COALESCE(org.fiscal_code, org_sp.fiscal_code) to correctly select the organizer from either the event (for tickets) or season pass price event (for season passes).

Issue #124, opened 2017-10-06
Find out why make_cover is not working


Ana06:
Check why TrelloWrapper is not working. It seems to work sometimes but not always, but I've tested always after calling `add_attachment` and it seems that with this method the attachment is also sometimes added as cover, so it may be that it is not working at all.

**Tip:** Consider stop using `put` call and use [Ruby Trello](https://github.com/jeremytregunna/ruby-trello) as mentioned here: https://github.com/openSUSE/trollolo/issues/62

Ana06:
It could be related with the fact that we upload several times an image with the same card to the card, as it seems it set the first one as cover.

PR #142, opened 2018-01-17
Use attach id instead of name for --plot-to-board

Ana06:
As it fails when having repeated names, which is normally the case.

Closes https://github.com/openSUSE/trollolo/issues/124


agraul (approved):
Looks good.

PR #153, opened 2018-02-01
Added option to select the directory for saving the png file.

### LINKED ISSUE #130: Being able to generate burndown chart image
Currently we have an option to select the working directory, where the data file is. At the moment the `png` with the burndown chart is in the same directory. We should implement a new option in `burndown` command, so that this directory can be different.

rishabhptr:
Fixes #130

### DIFF
--- a/lib/burndown_chart.rb
+++ b/lib/burndown_chart.rb
@@ -169,6 +169,7 @@ class BurndownChart
         cli_switches << '--with-fast-lane'           if hash['with-fast-lane']
         cli_switches << "--output #{hash['output']}" if hash['output']
         cli_switches << '--verbose'                  if hash['verbose']
+        cli_switches << "--location #{hash['location']}" if hash['location']
       end
     end
 
--- a/scripts/create_burndown.py
+++ b/scripts/create_burndown.py
@@ -26,11 +26,15 @@ def parseCommandLine():
   parser = argparse.ArgumentParser(epilog=epilog, description=description)
   parser.add_argument('sprint', metavar='NUM', help='Sprint Number')
   parser.add_argument('--output', help='Location of data to process')
+  parser.add_argument('--location', help='Image file location')
   parser.add_argument('--no-tasks', action='store_true', help='Disable Tasks line in the chart', default=False)
   parser.add_argument('--with-fast-lane', action='store_true', help='Draw line for Fast Lane cards', default=False)
   parser.add_argument('--verbose', action='store_true', help='Verbose Output', default=False)
   parser.add_argument('--no-head', action='store_true', help='Run in headless mode', default=False)
   args = parser.parse_args()
+  
+  if args.location:
+    args.location = os.path.join(os.getcwd(),args.location)
 
   if args.output:
     os.chdir(args.output)
@@ -118,7 +122,7 @@ if args.with_fast_lane:
 
   graph_fast_lane.draw(y_label, color, color_unplanned, marker, linestyle, linewidth, label, plot)
   legend_list.append(graph_fast_lane.subplot)
-    
+
 # Draw legend
 plot.drawLegend(legend_list)
 
--- a/scripts/plot.py
+++ b/scripts/plot.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import matplotlib.pyplot as plt
+import os
 
 
 class Plot:
@@ -68,9 +69,15 @@ class Plot:
     return
 
   def saveImage(self, args):
-    plt.savefig('burndown-' + args.sprint + '.png',bbox_inches='tight')
+    path = args.location
+    image_path = 'burndown-' + args.sprint + '.png'
+    if path:
+        if not os.path.exists(path):
+            os.makedirs(path)
+        image_path = os.path.join(path,image_path)
+    plt.savefig(image_path,bbox_inches='tight')
     if not args.no_head:
-      plt.show()
+        plt.show()
     return
 
   def createSubplot(self):
--- a/lib/cli/burndown.rb
+++ b/lib/cli/burndown.rb
@@ -21,6 +21,7 @@ class CliBurndown < Thor
   option :weekend_lines, type: :array, desc: 'Set the weekend_lines. [3.5, 8.5] by default'
   option :plot, type: :boolean, desc: 'also plot the new data'
   option :plot_to_board, type: :boolean, desc: 'Send the plotted data to the board'
+  option :location, desc: 'Image output directory', required: false
   option 'with-fast-lane', desc: 'Plot Fast Lane with new cards bars', required: false, type: :boolean
   option 'no-tasks', desc: 'Do not plot tasks line', required: false, type: :boolean
   option 'push-to-api', desc: 'Push collected data to api endpoint (in json)', required: false
--- a/man/trollolo.1.md
+++ b/man/trollolo.1.md
@@ -73,6 +73,11 @@ For correct generation of the burndown chart, the Trello board has to follow
 a few convention. They are described in the section `CONVENTIONS for SCRUM
 BOARDS`.
 
+`trollolo burndown update --output=<output_directory> --location=<location_directory> --plot`
+
+Update the burndown data in output_directory and plot the burndown chart for 
+the data and save the file in location_directory.
+
 ### burndown multi-update -- Process data for multiple burndown charts at once
 
 `trollolo burndown multi-update --board-list=<board list>`
--- a/spec/integration/create_burndown_spec.rb
+++ b/spec/integration/create_burndown_spec.rb
@@ -34,6 +34,15 @@ describe 'create_burndown.py' do
     end
   end
 
+  it 'creates image file in given location' do
+    @working_dir = given_directory do
+      given_file('burndown-data-01.yaml', from: 'create_burndown_helper/burndown-data-01.yaml')
+    end
+    result = run_helper(@working_dir, '01', ['--location=test'])
+    expect(result).to exit_with_success('')
+    expect(File.exist?(File.join(@working_dir, 'test/burndown-01.png'))).to be true
+  end
+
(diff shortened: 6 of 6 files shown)


rishabhptr:
@Ana06 Yes you are right :smile: . Actually it is because when we execute the python scripts we change directory to the output directory so I think if we store the current directory  user would be able to choose path relative to his current path. I have made the change.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 
 ## Version 0.2.0
 
+* Allow to select location to save burndown chart file through `--location`
```
Ana06: I would say **image** instead of file, to avoid confusion with the `.yml` file.

scripts/create_burndown.py
```diff
@@ -26,12 +26,16 @@ def parseCommandLine():
...
   parser.add_argument('--output', help='Location of data to process')
+  parser.add_argument('--location', help='Image file location')
   parser.add_argument('--no-tasks', action='store_true', help='Disable Tasks line in the chart', default=False)
   parser.add_argument('--with-fast-lane', action='store_true', help='Draw line for Fast Lane cards', default=False)
   parser.add_argument('--verbose', action='store_true', help='Verbose Output', default=False)
   parser.add_argument('--no-head', action='store_true', help='Run in headless mode', default=False)
   args = parser.parse_args()
 
+  if args.location:
+    args.location = os.path.join(os.getcwd(),args.location)
```
Ana06: what about doing this inside `saveImage`? It is actually not needed at this point. 😉

scripts/plot.py
```diff
@@ -68,7 +69,15 @@ def drawLegend(self, legend_list):
     return
 
   def saveImage(self, args):
-    plt.savefig('burndown-' + args.sprint + '.png',bbox_inches='tight')
+    path = args.location
+    filename = 'burndown' + args.sprint + '.png'
+    if path:
```
Ana06: you could write

```python
if args.location:
  path = os.path.join(os.getcwd(),args.location)
```

scripts/create_burndown.py
rishabhptr: Actually, when we pass --output=' ' for indicating the data file we switch to that directory at the very next line of code and saveImage is called in the last so when saveImage is called the current directory would be 'output' directory and not the directory the user is in now so I updated this here. :smile:

rishabhptr:
@Ana06 I have added the test but I encountered a issue because as with current functionality we open the image file automatically upon creation which would occur in running test too which isn't desired.

I noticed there is a option '--no-head' for this specific issue but it wasn't working as it was not present in cli.rb so I added it there and using it in my test.Please suggest if I should make some change in this.

lib/burndown_chart.rb
```diff
@@ -160,6 +160,7 @@ def process_options(hash)
         cli_switches << "--output #{hash['output']}" if hash['output']
         cli_switches << '--verbose'                  if hash['verbose']
         cli_switches << "--location #{hash['location']}" if hash['location']
+        cli_switches << '--no-head' if hash['no-head']
```
Ana06: you shouldn't modify the code to be able to write a test 😱

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 
 ## Version 0.2.0
 
+* Allow to select location to save burndown chart image through `--location`
```
Ana06: this should have been in the _Master (unreleased)_ section, but we have decided in https://github.com/openSUSE/trollolo/pull/163 to stop using this. So please remove this, it is not needed any more. And sorry for the mess. 😄

lib/burndown_chart.rb
rishabhptr: Sorry, actually I thought this could be implemented as a option and didnt realize that it had to be passed directly to python scripts for automation.

rishabhptr:
@Ana06 Thanks for the review, I have made some changes. I have re-written test to directly use no-head by passing it while executing python script.

Ana06:
@rishabhptr the number of contributors in Trollolo is increasing what is decreasing the velocity to review PR 😞  I'll take a look at soon as possible. 😉

rishabhptr:
@Ana06 No problem 😄 , and actually there is more work to it as earlier travis was showing error for not having python-matplotlib installed so I think we might need to add it in travis and also set up some display parameters to successfully execute the python script as directory and file will be created by them.

rishabhptr:
@Ana06 The tests are running fine locally, but on travis I think they need matplot installed to run the script so I think a change in travis configuration might be needed to test it properly.

.travis.yml
```diff
@@ -3,6 +3,9 @@ cache: bundler
 before_install:
   - gem --version
   - gem list bundler
+  - sudo apt-get install python-matplotlib
+  - "export DISPLAY=:99.0"
```
Ana06: why is this needed? :confused:

.travis.yml
```diff
@@ -3,6 +3,9 @@ cache: bundler
...
+  - "sh -e /etc/init.d/xvfb start"
```
Ana06: why is this needed? :confused:

spec/unit/burndown_chart_spec.rb
```diff
@@ -492,12 +492,21 @@
...
+
     it 'sends joined parsed options to python script' do
       allow(described_class).to receive(:process_options).and_return(%w{ --test 1 --no-blah })
       allow(described_class).to receive(:plot_helper).and_return('mescript')
       expect(described_class).to receive(:system).with('python mescript 42 --test 1 --no-blah')
       described_class.plot(42, foo: 1, bar: 2)
     end
+
+    it 'saves image file in given location' do
+      dir = given_directory
```
Ana06: why do you need this variable? :confused:

spec/unit/burndown_chart_spec.rb
```diff
@@ -492,12 +492,21 @@
...
+      allow(described_class).to receive(:process_options).and_return(%W{--output spec/data/burndown_dir --location #{dir} --no-head})
+      described_class.plot(2, foo: 1)
+      expect(File.exist?(File.join(dir, 'burndown02.png'))).to be true
```
Ana06: does this mean that test is creating the Image? Should we maybe delete it? I am not sure if that happens with other tests. :thinking:

spec/unit/burndown_chart_spec.rb
rishabhptr: It is actually the temporary directory created for writing the test data ( .png file ) ,as  we are using GivenFileSystemSpecHelpers in tests. 
http://www.rubydoc.info/gems/given_filesystem/0.1.2/file/README.md#Thanks

spec/unit/burndown_chart_spec.rb
rishabhptr: Yes, we are creating the image from our test data that already exists because then only we can test that image would be saved in the right place.
As ,I am mocking the parameters to point '--location' to the temp directory I dont think we need to do anything else as file wont exist after test run.

Also, I did this because this is how we are testing the functionality of '--output' option to  check that data file (.YAML) are saved in correct directory.

https://github.com/openSUSE/trollolo/blob/4f2df0497e6fa6e168eed7cc17d849335312ccda/spec/unit/cli_spec.rb#L14-L20

.travis.yml
rishabhptr: I had run travis on my fork and had faced issue of not having any display parameter for running python script so I used this here for headless GUI for testing.
https://docs.travis-ci.com/user/gui-and-headless-browsers/

cornelius: Does that mean that you are running matplotlib in the normal unit tests? I would avoid that because the setup can be tricky and strongly depends on the environment. This doesn't give results well defined enough to be reliably checked in tests.

We have the integration tests which run matplotlib in a container: `integration/create_burndown_spec.rb` This works very reliably.

spec/unit/burndown_chart_spec.rb
Ana06: but why do you need to assign it to a variable? :confused:

.travis.yml
Ana06: I agree with @cornelius :+1:

rishabhptr: Yes if we have a docker container for matplotlib then it should be used,I will make the changes and also sorry for late response as I was busy with exams past week. :smile:

Ana06:
@rishabhptr what is the state of this? do you need help? :smile:

rishabhptr:
@Ana06  As I mentioned earlier in comments I am sorry for late response due to exams. Also if we have docker container for matplotlib it was not required in travis :smile:  so I have made the changes.

rishabhptr:
@Ana06 Thanks for the reply :smile: . Sorry for the late response as I was away, I have rebased my branch but I'm encountering a syntax error in one of the python files called by --plot command. I am not really sure if it's same for others.

rishabhptr:
@Ana06 Yes the error has now resolved after rebase :smiley: . I have resolved the conflicts, please have a look.

scripts/plot.py
```diff
@@ -68,9 +69,17 @@ def drawLegend(self, legend_list):
     return
 
   def saveImage(self, args):
-    plt.savefig('burndown-' + args.sprint + '.png',bbox_inches='tight')
+    path = args.location
+    filename = 'burndown-' + args.sprint + '.png'
```
Ana06: I would name this variable path_image and add the path to it, so that you don't need the else, so:
```ruby
path = args.location
image_path = 'burndown-' + args.sprint + '.png'

if path
  if not os.path.exists(path):
    os.makedirs(path)
  image_path = os.path.join(path, image_path)

plt.savefig(image_path, bbox_inches='tight')

rishabhptr: Yeah this cleans up some code, thanks for the suggestion :smile:

rishabhptr:
@Ana06 Thanks for the review, I have made the changes :smile.

Issue #138, opened 2017-11-09
Label last sprint cards in cleanup-sprint


Ana06:
When executing `cleanup-sprint` the cards are moved to the backlog. It would be great to label then as `in the last sprint`. This should be in an option (`labelled=TRUE` for example), as may want that you don't want by default.

ankitkataria:
@Ana06 I wanted to contribute to trolllolo, can i claim this issue? also could i get some background on the project such as how to set it up? and does this project have documentation?

Ana06:
@ankitkataria all yours! 😉 

Trollolo is a quite simple Ruby gem and it should be difficult to set it up. That is why it does not much documentation, although if you think it would be useful to have more documentation, fell free to send a PR. 😉 

We have information about running the tests in https://github.com/openSUSE/trollolo/blob/master/CONTRIBUTING.md

Apart from that, you should only need to install the gems with bundle. To try it locally you can build it and install it, you can read how to do it for example here: http://guides.rubygems.org/make-your-own-gem/

ankitkataria:
on it :smile:

Ana06:
@ankitkataria only openSUSE biggest project have a dedicated IRC channel, until now there was not the need for an IRC channel for Trollolo. You can contact me directly in IRC (Ana06). 😉

ankitkataria:
@Ana06 I made a PR(https://github.com/openSUSE/trollolo/pull/156) i needed some help on managing the new labels, please check. :sweat_smile:

PR #156, opened 2018-02-03
Adds 'in the last sprint' label during cleanup-sprint

ankitkataria:
- Adds 'labelled' option in file cli.rb with default value of False
- Adds add_in_last_sprint_label method to add the required label to the moved card if labelled=True

Hey, @Ana06 this is my initial attempt at resolving #138 , What all I figured out: 
- Trollolo uses [thor] (https://github.com/erikhuda/thor/) to manage the cli interface, I added the labelled option to the cleanup_sprint function in cli.rb
- The variable, along with other arguments, is passed to cleanup method located in scrum/sprint_cleaner.rb
- I defined a add_in_last_sprint_label method to perform the addition of label and called it in move_cards if labelled is 'True'
- I referred [ruby-trello]documentation(http://www.rubydoc.info/gems/ruby-trello/Trello/Card#add_label-instance_method) to find out about the add_label method

I ran the test - `bundle exec rspec spec/unit/scrum/sprint_cleaner_spec.rb` and it threw me the following error, 

```
 ArgumentError:
       wrong number of arguments (given 2, expected 3)
     # ./lib/scrum/sprint_cleaner.rb:5:in `cleanup'
```

I did add the new option 'labelled' in cleanup_sprint in cli.rb, i've added a defualt as 'False' for it too, in the sprint_cleaner_spec file, only two arguments have been passed to test the cleanup function, shouldn't the default value of labelled reach in that case? I'm unable to understand why, Am i missing something? :thinking:  
Could I also get some help on how labels are managed in trollolo? how could i add a new one along with the others? the way i'm adding the label right now is incorrect.

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -349,13 +349,14 @@ EOT
 EOT
   option 'board-id', desc: 'Id of the board', required: true
   option 'target-board-id', desc: 'Id of the target board', required: true
+  option 'set-last-sprint-label', desc: 'Set true to label cards as - in the last sprint', required: false, type: :boolean, default: false
   def cleanup_sprint
     process_global_options options
     require_trello_credentials
 
     s = Scrum::SprintCleaner.new(@@settings)
     s.cleanup(board_id(options['board-id']),
-              board_id(options['target-board-id']))
+              board_id(options['target-board-id']), options['set-last-sprint-label'])
   end
 
   desc 'move-backlog', 'Move the planning backlog to the sprint board'
--- a/lib/scrum/sprint_cleaner.rb
+++ b/lib/scrum/sprint_cleaner.rb
@@ -2,7 +2,7 @@ module Scrum
   class SprintCleaner < TrelloService
     include ScrumBoards
 
-    def cleanup(board_id, target_board_id)
+    def cleanup(board_id, target_board_id, set_last_sprint_label: false)
       @board = sprint_board(board_id)
       raise "backlog list '#{@board.backlog_list_name}' not found on sprint board" unless @board.backlog_list
       @target_board = Trello::Board.find(target_board_id)
@@ -10,9 +10,9 @@ module Scrum
 
       gen_burndown
 
-      move_cards(@board.backlog_list)
-      move_cards(@board.doing_list) if @board.doing_list
-      move_cards(@board.qa_list) if @board.qa_list
+      move_cards(@board.backlog_list, set_last_sprint_label, target_board_id)
+      move_cards(@board.doing_list, set_last_sprint_label, target_board_id) if @board.doing_list
+      move_cards(@board.qa_list, set_last_sprint_label, target_board_id) if @board.qa_list
     end
 
     private
@@ -39,13 +39,19 @@ module Scrum
       card.remove_label(label) if label
     end
 
-    def move_cards(source_list)
+    def add_in_last_sprint_label(card, target_board_id)
+      @label = Trello::Label.create(name: @settings.scrum.label_names['is_last_sprint'], board_id: target_board_id)
+      card.add_label(@label)
+    end
+
+    def move_cards(source_list, set_last_sprint_label, target_board_id)
       source_list.cards.each do |card|
         next if @board.sticky?(card)
         puts %(moving card "#{card.name}" to list "#{target_list.name}")
         card.members.each { |member| card.remove_member(member) }
         remove_waterline_label(card)
         remove_unplanned_label(card)
+        add_in_last_sprint_label(card, target_board_id) if set_last_sprint_label
         card.move_to_board(@target_board, target_list)
       end
     end
--- a/lib/settings.rb
+++ b/lib/settings.rb
@@ -72,7 +72,8 @@ class Settings
       },
       'label_names' => {
         'sticky' => 'Sticky',
-        'waterline' => 'Under waterline'
+        'waterline' => 'Under waterline',
+        'in_last_sprint' => 'In last sprint'
       },
       'list_names' => {
         'sprint_backlog' => 'Sprint Backlog',
--- a/spec/unit/scrum/sprint_cleaner_spec.rb
+++ b/spec/unit/scrum/sprint_cleaner_spec.rb
@@ -7,9 +7,19 @@ describe Scrum::SprintCleaner do
     expect(subject).to be
   end
 
-  it 'moves remaining cards to target board', vcr: 'sprint_cleanup', vcr_record: false do
-    expect(STDOUT).to receive(:puts).exactly(13).times
-    expect(subject.cleanup('7Zar7bNm', '72tOJsGS')).to be
+  context 'given set-last-sprint-label flag is true' do
+    before do
+      allow(Trello::Label).to receive(:create)
+      allow_any_instance_of(Trello::Card).to receive(:add_label)
+    end
+
+    it 'moves remaining cards to target board', vcr: 'sprint_cleanup', vcr_record: false do
+      expect(STDOUT).to receive(:puts).exactly(13).times
+      expect(subject.cleanup('7Zar7bNm', '72tOJsGS', set_last_sprint_label: true)).to be
+      subject.sprint_board('7Zar7bNm').backlog_list.cards.each do |card|
+        expect(card).to receive(:add_label).once
+      end
+    end
   end
 
   context 'given correct burndown-data-xx.yaml' do


lib/cli.rb
```diff
@@ -337,13 +337,14 @@ def set_priorities
 EOT
   option 'board-id', desc: 'Id of the board', required: true
   option 'target-board-id', desc: 'Id of the target board', required: true
+  option 'labelled', desc: 'Set true to label cards as - in the the last sprint', default: 'False'
```
Ana06: This should be of type `boolean`. Check [thor documentation](https://github.com/erikhuda/thor/wiki/Method-Options#types-for-method_options) for details.

ankitkataria:
> You need to update the test. The method you modified it is called from there and it has only two arguments, that why it fails. You need to give it an extra argument.

Oh I was thinking that after specifying the default value, and setting required: false, the third value was not actually required while calling the 

> I am not sure what you mean, can you please elaborate? 

I wanted to know how I could make and add cards, i figured it out after reading [ruby-trello docs](http://www.rubydoc.info/gems/ruby-trello/Trello/Label)

> And please fix the style offenses, Rubocop is unhappy

Sorry about that :sweat_smile: I'll take care of that from now on.

Ana06:
@ankitkataria 

> Oh I was thinking that after specifying the default value, and setting required: false, the third value was not actually required while calling the

it is not if you call the option from the console, but the test is directly calling the method.

ankitkataria:
@Ana06, I had a doubt, I was making a call to 
```
@label=Trello::Label.create(name:@settings.scrum['in_last_sprint']) 
card.card_labels << under_in_last_sprint_label
```
I was doing this to make a new label and add it to the card that was being moved. but the first line threw me a, `VCR::Errors::UnhandledHTTPRequestError:`. I changed the vcr record mode to return :new_episodes instead of :none in /spec/unit/support/vcr.rb so that it could log the new requests too, but after that the Label.create statement gave me a `Invalid Key Error`. I don't know how to proceed.

I've pushed all the latest changes that i've made, i'm a little new to the use of rspec and vcr and am learning about them. I'll  try to figure them out during the weekend, i'll be grateful for any help meanwhile :smile: 

PS. I tried reaching you through IRC, but I'm unsure if you got the message, hence posting here.

Ana06:
@ankitkataria what is the state of this? do you need help?

ankitkataria:
hey @Ana06, i had already tried adding the board it, it still threw me the invalid key error, i was wondering if i had to initialize the test file with some dummy trello key, to eliminate the error. :thinking: 

image

ankitkataria:
@Ana06 any help possible here?
will it be okay if i start working on some other issue too side by side?

Ana06:
@ankitkataria

> @Ana06 any help possible here?

Sorry the delay, but the number of contributors in Trollolo is increasing what is decreasing the velocity to review PR 😞  I'll take a look at soon as possible. 😉 

> will it be okay if i start working on some other issue too side by side?

@ankitkataria it is not forbidden that you start working on something else, but I think it is better not to work on many things at the same time. There are also other things that you can do that will help you to work on Trollolo apart from sending PR, such as reading about scrum, reading trello developer documentation, etc... If you need help to find resources I can also point you to some material. 😉

Ana06:
@ankitkataria what is the state of this? do you need help?

Ana06:
@ankitkataria ^^

ankitkataria:
Hey @Ana06, Sorry haven't been able to work on this for a long time. Honestly, I still don't know, where I'm lacking. :disappointed:

Ana06:
@ankitkataria have you tried this in a real environment apart that in the test suit? does it work there? :thinking:

ankitkataria:
@Ana06 , figured what had been troubling me :sweat:. I apologize for having disappeared. I hope we can still work on this PR and get it merged.

lib/cli.rb
```diff
@@ -349,13 +349,14 @@ def set_priorities
 EOT
   option 'board-id', desc: 'Id of the board', required: true
   option 'target-board-id', desc: 'Id of the target board', required: true
+  option 'labelled', desc: 'Set true to label cards as - in the the last sprint', required: false, type: :boolean, default: false
```
Ana06: `the` in the description is twice. And it is not really clear, what about this for the desc: `If true cards are label as "In last sprint"`

lib/scrum/sprint_cleaner.rb
```diff
@@ -39,13 +39,19 @@ def remove_unplanned_label(card)
       card.remove_label(label) if label
     end
 
-    def move_cards(source_list)
+    def add_in_last_sprint_label(card, target_board_id)
+      @label = Trello::Label.create(name: 'In Last Spring', board_id: target_board_id)
```
Ana06: you should be using the name from the settings so that it can be customized, similarly as it is done with the waterline label: `@settings.label_names['in_last_sprint']`

lib/settings.rb
```diff
@@ -72,7 +72,8 @@ def scrum_defaults
       },
       'label_names' => {
         'sticky' => 'Sticky',
-        'waterline' => 'Under waterline'
+        'waterline' => 'Under waterline',
+        'in_last_sprint' => 'In Last Sprint'
```
Ana06: To be consistent with the other labels I think we shouldn't capitalise every word, so: `'In last sprint'`

spec/unit/scrum/sprint_cleaner_spec.rb
```diff
@@ -8,8 +8,9 @@
   end
 
   it 'moves remaining cards to target board', vcr: 'sprint_cleanup', vcr_record: false do
+    allow_any_instance_of(Trello::Label).to receive(:create)
     expect(STDOUT).to receive(:puts).exactly(13).times
-    expect(subject.cleanup('7Zar7bNm', '72tOJsGS')).to be
+    expect(subject.cleanup('7Zar7bNm', '72tOJsGS', true)).to be
```
Ana06: How is that testing that the cards are labelled? :thinking:

spec/unit/scrum/sprint_cleaner_spec.rb
```diff
@@ -8,8 +8,10 @@
   end
 
   it 'moves remaining cards to target board', vcr: 'sprint_cleanup', vcr_record: false do
-    expect(STDOUT).to receive(:puts).exactly(13).times
-    expect(subject.cleanup('7Zar7bNm', '72tOJsGS')).to be
+    allow(Trello::Label).to receive(:create)
+    allow_any_instance_of(Trello::Card).to receive(:add_label).and_return(true)
+    expect(STDOUT).to receive(:puts).exactly(25).times
```
ankitkataria: @Ana06 I increased the check from the previous value 13 to 25. I added a puts statement in [add_in_last_sprint_label](https://github.com/openSUSE/trollolo/pull/156/files#diff-eecdf5816db86046bb09563f62a60f0cR44) That amounts for the extra 12. Also now I'm using  [`add_label`](https://github.com/jeremytregunna/ruby-trello/blob/master/lib/trello/card.rb#L402) method of the library to add our new label, whose call I've stubbed in the tests.

Ana06: I don't really like to modify the code to make a test passing...

Ana06: and the comment is not really testing that the card label is added...

spec/unit/scrum/sprint_cleaner_spec.rb
```diff
@@ -7,9 +7,19 @@
...
+  context 'given labelled flag is true' do
+    before do
+      allow(Trello::Label).to receive(:create)
+      allow_any_instance_of(Trello::Card).to receive(:add_label)
+    end
+
+    it 'moves remaining cards to target board', vcr: 'sprint_cleanup', vcr_record: false do
+      expect(STDOUT).to receive(:puts).exactly(13).times
+      expect(subject.cleanup('7Zar7bNm', '72tOJsGS', true)).to be
+      subject.sprint_board('7Zar7bNm').backlog_list.cards.each do |card|
```
ankitkataria: does this work? now we're expecting all the instances of card to receive the add label call which actually adds the labels.

ankitkataria: @Ana06 ?

lib/scrum/sprint_cleaner.rb
```diff
@@ -2,17 +2,17 @@ module Scrum
   class SprintCleaner < TrelloService
     include ScrumBoards
 
-    def cleanup(board_id, target_board_id)
+    def cleanup(board_id, target_board_id, labelled)
```
cornelius: Please use a named parameter instead of a boolean. When you read a call such as `cleanup(some_id, other_id, false)` you don't know what the `false` is doing.

lib/cli.rb
```diff
@@ -349,13 +349,14 @@ def set_priorities
 EOT
   option 'board-id', desc: 'Id of the board', required: true
   option 'target-board-id', desc: 'Id of the target board', required: true
+  option 'labelled', desc: 'Set true to label cards as - in the last sprint', required: false, type: :boolean, default: false
```
cornelius: I would suggest to use another name for the option. `labelled` sounds like the label is already there. I would rather express it as an action. It also would be good to make it clear what label is added. Maybe `--set-last-sprint-label`?

ankitkataria:
@cornelius moved to named parameter and changed the label name :smile:

cornelius:
One general comment about pull requests: I would never force push a branch of a pull request. Adding changes as additional commits makes it much more easy to follow the flow of the discussion and review changes. The branch can still be squashed into one commit or a smaller series of commits when merging after review.

ankitkataria:
Thankyou @cornelius :smile: . Sorry about that, won't do while contributing again. @Ana06 are there any other improvements?

Ana06:
@cornelius 

> I would never force push a branch of a pull request. Adding changes as additional commits makes it much more easy to follow the flow of the discussion and review changes

I actually would do it the other way around (it may be that I told @ankitkataria to do it like that at some point) because IMO a clean commit history helps the review, even if it involves reviewing old changes. But if @cornelius finds it easier to add new commits and squash them before merging, I can live with it too. :sweat_smile:

ankitkataria:
Thanks @Ana06! :tada:

PR #161, opened 2018-02-10
Remove redundant gem --version

Ana06:
`gem --version` is automatically shown, so there is no need to duplicate it.

image


cornelius (approved):
Looks good to me

PR #167, opened 2018-02-16
Add rake task for style and tests

### LINKED ISSUE #164: Rake task for test
As suggested by @cornelius in https://github.com/openSUSE/trollolo/pull/163, it would be nice to have a rake task to run the tests and style checkers, so that everything is run together and there is not need of remember the options.

It should be used in Travis as well.

rahul2240:
Instead of remembering all the commands it would be nice to have rake task to run tests and style checkers
closes https://github.com/openSUSE/trollolo/issues/164

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,3 +10,20 @@ namespace :gem do
     system 'gem build trollolo.gemspec'
   end
 end
+
+require 'rubocop/rake_task'
+require 'rspec/core/rake_task'
+
+task default: %i[rubocop spec]
+
+desc 'Run tests'
+task :spec do
+  RSpec::Core::RakeTask.new
+end
+
+desc 'Run rubocop'
+task :rubocop do
+  RuboCop::RakeTask.new do |t|
+    t.options = ['--display-cop-names']
+  end
+end


Ana06:
@rahul2240 there is quite a lot of PRs in Trollolo at the moment, so it may take some days until I have time to review this. 😉

Rakefile
```diff
@@ -10,3 +10,20 @@ namespace :gem do
     system 'gem build trollolo.gemspec'
   end
 end
+
+# rake task for rspec and rubocop
+# rake -T (for rake tasks)
```
Ana06: why is this comment needed here? 😕

Rakefile
```diff
@@ -10,3 +10,20 @@ namespace :gem do
...
+require 'rubocop/rake_task'
+require 'rspec/core/rake_task'
+
+task default: %i[rubocop spec]
+
+desc 'Run tests'
```
Ana06: @rahul2240 the issue is about creating a single rake task that run both the tests and rubocop. 😉

Rakefile
```diff
@@ -10,3 +10,20 @@ namespace :gem do
...
+task :spec do
+  RSpec::Core::RakeTask.new
+end
+
+desc 'Run rubocop'
+task :rubocop do
+  RuboCop::RakeTask.new
```
Ana06: this is not including the same options documented in [the contributing guide](https://github.com/openSUSE/trollolo/blob/master/CONTRIBUTING.md) and used in Travis. It should be the same.

Ana06 (changes requested):
@rahul2240 thanks for your PR. 💐  I think you misunderstood the issue. It is about having a single task that run what is run in Travis 😉

Rakefile
rahul2240: I will remove it.

Rakefile
rahul2240: @Ana06 using rake will do both the tasks (test,style check) if there is some offence then only it will not complete.

Ana06: @rahul2240 what do you mean with `using rake`? what command would you run?

rahul2240: using "rake" only will do both the task.

Ana06: @rahul2240 and what happens if we add another unrelated rake task in the future?

rahul2240: @Ana06 Right now test and style check is set as default for rake command if unrelated rake task is added in future then we will have to use (rake taskname) to run it.

rahul2240: @Ana06 If I am doing it wrong please let me know I am ready to start again :smile:

Ana06: @rahul2240 if we add a task in the future you will have to run rake taskname and then you can not run both things at the same time. We want a single rake task, not two, that runs everything that Travis run.

rahul2240: @Ana06 Ok now I got it. Cann't we add that task to default? so that it runs automatically with rake command, Or we need a more simpler method?

Ana06 (changes requested):
You should not add them to the default. As already said  we want a single rake task, that runs everything that Travis run.

Ana06:
@rahul2240 what is the state of this? do you need help?

rahul2240:
@Ana06 I am searching for the solution, give me some more time please.

rahul2240:
@Ana06  We can use namespace and add all the rake task there or in my code instead of adding it to default we can name it as "check" so that  (rake check) command will run all the test.

Ana06:
@rahul2240 what about doing it how [I suggested already](https://github.com/openSUSE/trollolo/pull/167#discussion_r170197374)? **We want a single rake task that runs everything that Travis run.** :wink: The namespace would also work, but you don't need to complicate that much :smile:

rahul2240:
@Ana06 Travis runs 2 tasks i.e 'bundle exec rubocop -D' &  'bundle exec rspec spec/unit' so what you want is in .rakefile i remove the rake tasks and target travis.yml to run those tasks locally with a single command?

Issue #45, opened 2016-10-26
Command to show backlog


cornelius:
When doing sprint commitments we need a way to see the whole backlog in one view.

To achieve that add a `show-backlog` command which lists the backlog on the terminal in a table. The table should have one column for the priority, one for the size and one for the title. 

To show the average velocity we should add a command line option `--velocity` which specifies an average velocity. This adds a line in the table at the location where the sum of the cards before is closest to the specified velocity.

jimmykarily:
Is this the expected output format?:

```
$ bin/trollolo show-backlog --board-id=Test

Priority | Points | Title
       3 |      7 | P3: (7) Dress like a superhero
       5 |      2 | P5: (2) Save the world
       6 |      4 | P6: (4) Celebrate
```

I think having the first two columns is redundant since they are necessarily extracted from the card title. Maybe just dump the card names in priority order?

cornelius:
@jimmykarily You are right. The first two columns are redundant. Just showing the card names in priority order would be enough.

matheussbernardo:
I'm currently working on this issue, @cornelius @Ana06.
My problem is how to get the column in a properly way what i am doing today is.
`def show_backlog
    puts "Title"
    TrelloWrapper.new(@@settings).board(options['board-id']).open_columns[0].cards.each do |card|
      puts "| " + card.name
    end
  end`

getting the first column of the array of open columns but it seems like a bad solution!

matheussbernardo:
@cornelius The --velocity option should have this output

$ bin/trollolo show-backlog --board-id=Test --velocity 9

| Title
| P3: (7) Dress like a superhero
| P5: (2) Save the world
| (9) Velocity |
| P6: (4) Celebrate

Ana06:
@matheussbernardo you can send a PR to implement this without the velocity, and then another one for the velocity. This two thing can be done independently, so let's focus first on the `show-backlog` command to avoid having a too complicate and big PR. 😉

cornelius:
@matheussbernardo I agree with @Ana06 about separating the two parts in two different PRs. For the velocity option I would separate the code for handling the actual command line option such as parsing the option, putting out help, etc (which belongs in cli.rb and is mostly handled by the CLI parsing framework) and the "business logic" code which actually reads the data and prints out the results. The latter could be in a separate class (or maybe there is one where it would belong naturally). This also would make testing easier, because you can write simple and expressive unit tests which just call a method and don't need to provide input in the form of command line option or some derivation of that.

PR #168, opened 2018-02-16
Add show-backlog command

matheussbernardo:
This PR starts to work on #45,

I've created a method on ScrumBoard to get the  Sprint Backlog Column and then on the CLI I created the command and printed the information.

The --velocity option on the #45  is WIP.

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -215,6 +215,18 @@ EOT
     end
   end
 
+  desc 'show-backlog', 'Show backlog of board'
+  option 'board-id', desc: 'Id of Trello board', required: true
+  def show_backlog
+    puts '| Title'
+
+    trello = TrelloWrapper.new(@@settings)
+    board = trello.board(options['board-id'])
+    board.planning_backlog_column.cards.each do |card|
+      puts "| #{card.name}"
+    end
+  end
+
   desc 'show-backup', 'Show backup of board'
   option 'board-id', desc: 'Id of Trello board', required: true
   option 'show-descriptions', desc: 'Show descriptions of cards', required: false, type: :boolean
--- a/lib/scrum_board.rb
+++ b/lib/scrum_board.rb
@@ -12,6 +12,10 @@ class ScrumBoard
     @columns ||= @board_data['lists'].map{|x| Column.new(@board_data, x['id'], @settings)}
   end
 
+  def planning_backlog_column
+    columns.select{ |column| column.name == @settings.scrum['list_names']['planning_backlog'] }.first
+  end
+
   def todo_columns
     columns.select{|c| @settings.todo_columns.include?(c.name)}
   end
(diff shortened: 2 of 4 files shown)


lib/cli.rb
```diff
@@ -215,6 +215,15 @@ def list_backups
     end
   end
 
+  desc 'show-backlog', 'Show backlog of board'
+  option 'board-id', desc: 'Id of Trello board', required: true
+  def show_backlog
+    puts 'Title'
+    TrelloWrapper.new(@@settings).board(options['board-id']).sprint_backlog_column.cards.each do |card|
```
Ana06: if I am not wrong, this should be `planning_backlog_column` and not `sprint_backlog_column`.

@cornelius wrote in https://github.com/openSUSE/trollolo/issues/45#issue-185365473:

> When doing sprint commitments we need a way to see the whole backlog in one view.

Normally when preparing the next Sprint in a team using Scrum, you have to decide how many task the team is able to do in the Sprint. That why I think this is in the planning board and not in the sprint one. @cornelius is that right?

**Note:** A sprint can be for example two week. The team decide at the start what they are going to work in this two week. @matheussbernardo if you think it would be useful to read more about how scrum works, please tell me and I will send you some material. 😉

lib/cli.rb
```diff
@@ -215,6 +215,15 @@ def list_backups
     end
   end
 
+  desc 'show-backlog', 'Show backlog of board'
+  option 'board-id', desc: 'Id of Trello board', required: true
+  def show_backlog
+    puts 'Title'
```
Ana06: `puts '| Title'` 😉

lib/cli.rb
Ana06: Also, what about writing this like:

```ruby
trello = TrelloWrapper.new(@@settings)
board = trello.board(board_id(options['board-id']))


board.planning_backlog_column.cards.each do |card|
```

More readable when splitting it 😉

lib/scrum_board.rb
```diff
@@ -12,6 +12,10 @@ def columns
     @columns ||= @board_data['lists'].map{|x| Column.new(@board_data, x['id'], @settings)}
   end
 
+  def sprint_backlog_column
+    columns.select{ |col| col.name == 'Sprint Backlog' }.first
```
Ana06: I would prefer to write `column` instead of `col`, more explicit 😉 

The column name can be configured, as you can see in the [Documentation](https://github.com/openSUSE/trollolo#default-configuration), so you should use `settings.list_names['planning_backlog']` 😄

matheussbernardo: Hello @Ana06, the method list_names does not exist in the setting object, what I did was `@settings.scrum['list_names']['planning_backlog']`

cornelius:
Yes, this is about the backlog on the planning board.

matheussbernardo:
@Ana06 I updated as requested.

Ana06:
@matheussbernardo do you need help? The code is already great, we just need a test and we can :ship: it! :wink:

matheussbernardo:
@Ana06 Sorry for the delay, I was on vacation! I added the requested spec. Is there anything else to do?

matheussbernardo:
@Ana06  Is the squash correct? I`m having a hard time with git rebase -i.

Ana06:
@matheussbernardo

> @Ana06 Is the squash correct? I`m having a hard time with git rebase -i.

It doesn't seem so :sob: `rebase -i` is always tricky the first times :wink: So, first of all make a copy pf your changes, in case you destroy everything by accident.

Try with:

```
git remote add upstream https://github.com/openSUSE/trollolo.git
git fetch upstream
git rebase upstream/master
```

And then push it again. This should remove the current commits that are in your PR and that are not yours.

In case that this doesn't work and you don't manage to solve it, you can also ping me in IRC on Freenode (normally my nick is @Ana06, although sometimes it has a random number of 0s). Maybe easier :wink:.

matheussbernardo:
@Ana06  It worked! Hope we can 🚢 it now!

Issue #139, opened 2017-12-12
Trollolo doesn't work with Ruby 2.4


Ana06:
Everything seems to work in Ruby 2.4.1, but no wit my current 2.4.2:

```
> ruby -v
ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux-gnu]
```

That's the error:

```
> trollolo burndown --plot-to-board

/usr/lib64/ruby/gems/2.4.0/gems/rest-client-1.8.0/lib/restclient/request.rb:163:in `fetch': key not found: :ciphers (KeyError)
        from /usr/lib64/ruby/gems/2.4.0/gems/rest-client-1.8.0/lib/restclient/request.rb:163:in `initialize'
        from /usr/lib64/ruby/gems/2.4.0/gems/rest-client-1.8.0/lib/restclient/request.rb:41:in `new'
        from /usr/lib64/ruby/gems/2.4.0/gems/rest-client-1.8.0/lib/restclient/request.rb:41:in `execute'
        from /usr/lib64/ruby/gems/2.4.0/gems/ruby-trello-1.5.1/lib/trello/net.rb:29:in `execute_core'
        from /usr/lib64/ruby/gems/2.4.0/gems/ruby-trello-1.5.1/lib/trello/net.rb:18:in `try_execute'
        from /usr/lib64/ruby/gems/2.4.0/gems/ruby-trello-1.5.1/lib/trello/net.rb:10:in `execute'
        from /usr/lib64/ruby/gems/2.4.0/gems/ruby-trello-1.5.1/lib/trello/client.rb:88:in `invoke_verb'
        from /usr/lib64/ruby/gems/2.4.0/gems/ruby-trello-1.5.1/lib/trello/client.rb:19:in `get'
        from /usr/lib64/ruby/gems/2.4.0/gems/trollolo-0.1.1/lib/trello_wrapper.rb:84:in `get_board'
        from /usr/lib64/ruby/gems/2.4.0/gems/trollolo-0.1.1/lib/trello_wrapper.rb:32:in `retrieve_board_data'
        from /usr/lib64/ruby/gems/2.4.0/gems/trollolo-0.1.1/lib/trello_wrapper.rb:21:in `board'
        from /usr/lib64/ruby/gems/2.4.0/gems/trollolo-0.1.1/lib/burndown_data.rb:96:in `board'
[...]

PR #175, opened 2018-02-27
Update ruby-trello

Ana06:
Trollolo doesn't work with Ruby 2.4 because of using an old version of ruby-trello. :bowtie: 

Fixes https://github.com/openSUSE/trollolo/issues/139.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ before_install:
 rvm:
   - "2.2.3"
   - "2.3.1"
+  - "2.4.3"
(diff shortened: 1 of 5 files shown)


Ana06:
**DO NOT MERGE** - I need to check that the test failures do not mean that the code doesn't work.

Ana06:
[P12: (3) Frontpage: Burndown Chart](https://trello.com/c/Gm0FGST1/567-p12-3-frontpage-burndown-chart)

PR #184, opened 2018-03-01
Add --plot-to-board to plot command

### LINKED ISSUE #134: Add --plot-to-board to plot command
This option would be also useful in this command. This will send the plotted burndown chart to the first card of the `Done` column.

Ronaq13:
This flag of plot command will send the plotted burndown chart to the first
card of the `Done` column.

close #134

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -226,6 +226,18 @@ EOT
     end
   end
 
+  desc 'show-backlog', 'Show backlog of board'
+  option 'board-id', desc: 'Id of Trello board', required: true
+  def show_backlog
+    puts '| Title'
+
+    trello = TrelloWrapper.new(@@settings)
+    board = trello.board(options['board-id'])
+    board.planning_backlog_column.cards.each do |card|
+      puts "| #{card.name}"
+    end
+  end
+
   desc 'show-backup', 'Show backup of board'
   option 'board-id', desc: 'Id of Trello board', required: true
   option 'show-descriptions', desc: 'Show descriptions of cards', required: false, type: :boolean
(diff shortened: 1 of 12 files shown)


lib/cli.rb
```diff
@@ -189,13 +189,24 @@ def burndown
...
   option 'with-fast-lane', desc: 'Plot Fast Lane with new cards bars', required: false, type: :boolean
   option 'no-tasks', desc: 'Do not plot tasks line', required: false, type: :boolean
+  option :plot_to_board, type: :boolean, desc: 'Send the plotted data to the first card of Done column'
   def plot(sprint_number)
     process_global_options options
-    BurndownChart.plot(sprint_number, options)
+    if options[:plot_to_board]
+      begin
+        chart = BurndownChart.new @@settings
+        chart.update(options)
```
Ana06: This is also modifying the `yml` file, which we don't want to modify here. We only want to print the image and push it to the card.

Ana06 (changes requested):

lib/cli.rb
Ronaq13: Please help identifying the line at which it modifying the .`yml` file. And by .`yml` file do you mean burndown-data-01.yml file which is created at time of initialization

Ana06: inside of the `update` method.

Yes, the `burndown-data-i.yaml` file, where i is the number of the sprint

Ronaq13: @Ana06 I dont see any change in the yml file. Please check now is it good to merge?

Ana06:
@Ronaq13 @matheussbernardo @cornelius where are those commits coming from?

Ana06:
@Ronaq13 what is the state of this PR? :thinking:

Ana06:
Closing due to inactivity, @Ronaq13 feel free to reopen it if you continue working on it :wink:

PR #187, opened 2018-03-02
Rubocop offense

### LINKED ISSUE #151: Solve Style/MixinUsage RuboCop cop offenses
Solve `Style/MixinUsage` RuboCop cop offenses, currently excluded in the [*rubocop_todo* file](https://github.com/openSUSE/trollolo/blob/master/.rubocop_todo.yml). This cop checks that `include`, `extend` and `prepend` statements appear inside classes and modules, not at the top level, so as to not affect the behavior of `Object`. More info: https://rubocop.readthedocs.io/en/latest/cops_style/#stylemixinusage

Ronaq13:
Solved Style/MixinUsage RuboCop cop offenses, currently excluded in the rubocop_todo file.

close #151


spec/integration/command_line_spec.rb
```diff
@@ -1,52 +1,54 @@
 require_relative 'integration_spec_helper'
 
-include GivenFilesystemSpecHelpers
-include CliTester
+module CommandLineSpec
```
Ana06: This is an RSpec test not a module! :see_no_evil:

spec/integration/create_burndown_spec.rb
```diff
@@ -1,65 +1,67 @@
 require_relative 'integration_spec_helper'
 
-include GivenFilesystemSpecHelpers
-include CliTester
+module CreateBurndownSpec
```
Ana06: This is an RSpec test not a module! :see_no_evil:  and the same for the other RSpec files...

spec/unit/backup_spec.rb
```diff
@@ -1,40 +1,42 @@
 require_relative 'spec_helper'
+# rubocop:disable ModuleLength
```
Ana06: you shouldn't introduce new offenses while solve an offense... :disappointed:

Ronaq13: The module size is too large according to Rubocop. I thought the code logic needs to be in one file that's why I didn't split them.
In order to do that, I think I need to split this big module in small ones? Or we can do something as below in `rubocop.yml` :
`
ClassLength:
  Max: 250
`
 What you think on that?

spec/integration/command_line_spec.rb
Ronaq13: I am not clear about this. Please clarify, what I have done wrong.

spec/unit/backup_spec.rb
Ana06: you should not introduce new offenses, neither change the configuration. If a module is too long, some thing needs to be out of the module. But how you are fixing the `Style/MixinUsage` offenses is not right. Maybe you don't need to touch this, focus on solving `Style/MixinUsage` first. :wink:

Ronaq13:
@Ana06 checks aren't getting complete.

Ana06:
@Ronaq13 

> @Ana06 checks aren't getting complete.

It seems Travis got blocked. Rebase to master (as you have conflicts anyway) and push again. This should retrigger Travis. :wink:

Ana06:
@Ronaq13 will you finish this? do you need help to do it? If not I'll close it to allow that someone else work on it :wink:

Ronaq13:
I just forgot about this. I will push new changes in a day or two

Ronaq13:
Checks are getting passed locally. Don't know why it is not passing here. Should I create a new PR?

PR #190, opened 2018-03-06
Changed done_column_name_regex to done_columns

### LINKED ISSUE #189: Getting DoneColumnNotFoundError
I am following https://github.com/openSUSE/trollolo/blob/master/man/trollolo.1.md and have also setup `.trollolorc` file. 
My `.trollolorc` file looks like this :

screen shot 2018-03-06 at 3 32 19 am

And my board looks like :
screen shot 2018-03-06 at 4 09 25 am

But when i run `trollolo burndown` i am getting the following error:
screen shot 2018-03-06 at 9 10 37 pm

namangupta01:
Fixes #189 
Currently user have to write regex expression in .trollolorc file and even after writing regex expression done_column_name_regex was coming as an array of string but we need regex expression there so here array is converted into regex expression on the spot this also allows to write done_columns easily in the .trollolorc file

### DIFF
--- a/lib/scrum_board.rb
+++ b/lib/scrum_board.rb
@@ -25,11 +25,12 @@ class ScrumBoard
   end
 
   def done_column
-    done_columns = columns.select{|c| c.name =~ @settings.done_column_name_regex }
+    done_column_name_regex = Regexp.new(@settings.done_columns.join("|"))
+    done_columns = columns.select{|c| c.name =~ done_column_name_regex }
     if done_columns.empty?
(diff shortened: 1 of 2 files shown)


Ana06:
@namangupta01 it needs to be a regular expression because normally you have several Done columns with the number of the Sprint: `Done 1`, `Done 2`, `Done 3`,... or `Done Sprint 1`, `Done Sprint 2`, `Done Sprint 2`,...

Issue #119, opened 2017-10-06
Test suite need some love


Ana06:
We need to think about how to improve the test suite because currently there are many thing untested or test that are not actually testing what we want and are difficult to modify when changing the code.

Ana06:
I think that should be included in release 1.0.0. It is quite a lot of work :see_no_evil:

manno:
I noticed we have three(?) different approaches to dealing with the outgoing Trello API requests.
 
I'll just put my notes here:

* VCR tests: hard to write good expectations without exposing `board` and recording more requests since everything is a request? Regenerating requires example data on trello. Even though it's documented creating new VCRs looks complicated. Responses are just as real as in production.
* BoardMock: verbose, exposes `board` for no reason beside testing? Strong expectations on mocked data possible.
* `stub_requests`: only testing outgoing http, expectations even shallower than VCR. Seems like the oldest approach.

manno:
I came up with a branch to remove VCR from the Scrum tests: https://github.com/openSUSE/trollolo/tree/wip-remove-vcr

I'd like to get rid of them, because expectations were weak and when using them to iterate on stories they add manual work. I constantly have to watch out for outdated cassettes and switch between real and dummy settings.

PR #193, opened 2018-05-19
Wip remove vcr

manno:
This get's rid of VCR completely. Might still make sense to re-introduce it for integration tests, if we add more integration tests in the future.
I think it improves testing and helps with issue #119 

Instead of mocking outgoing http request, this mocks the dependencies, i.e. `Trello` from the trello gem, or if possible the `scrum/board*` classes.


Ana06:
@manno what is the state of this? The PR is quite big and there are already files conflicting... what is missed?

manno:
I created an updated PR #200

Issue #15, opened 2015-03-13
Restructure commands


cornelius:
The current command structure is a bit ad-hoc and has grown over time. We should clean them up. Here is a proposal @kalabiyau and me came up with:

`backup get`
`backup list`
`backup show`

`burndown init`
`burndown get`
`burndown plot`

`get [--raw] lists`
`get [--raw] cards`
`get [--raw] checklists`
`get [--raw] url <url-fragment>`

`help`

Remove the `fetch-burndown-data` command.

guilleiguaran:
@cornelius hey, I'm working on this and there are some other commands not included in the list and that should be moved to a subcommand (or keep as are currently), those are:

`organization`
`organization-members`
`move-backlog`
`sprint-cleanup`
`list-member-boards`

Also I suggest to move the next commands under the `set` subcommand:

`set priority`
`set cover`
`set description`

what do you think?

cornelius:
I would group the organization commands under one common sub command:

`organization show`
`organization members`

The `list-member-boards` is a special command more or less getting raw data, so this would fit well under the `get` commands:

`get [--raw] boards --member-id=<id>`

The `move-backlog` and `sprint-cleanup` commands are tied to a specific Scrum work flow, so I would group them under a `scrum` command:

`scrum prioritize` (was `set-priorities`)
`scrum start` (was `move-backlog`)
`scrum end` (was `cleanup-sprint`)
`scrum init` (was `setup-scrum`)

`set cover` and `set description` make sense to me.

matheussbernardo:
Great, I really liked the changes proposed.

So the changes are:

`burndown multi-update`
`get members`
`scrum backlog`
`scrum sprint`  *maybe it is not very clear that this is the sprint backlog*
and the --raw option is kept.

I would like to know if I can start to implement the changes. @Ana06 @cornelius

cornelius:
Let's leave out the `scrum sprint` for now, if it's not available yet anyway.

@matheussbernardo please go ahead with the implementation. We can discuss more details when reviewing your pull request.

PR #195, opened 2018-06-04
Restructure commands: Adding get subcommands (WIP)

matheussbernardo:
Hello @Ana06 @cornelius,

I started to implement the new commands, 
I faced a big problem with the tests. I do not fully understand them, It does not seem to have a pattern between all the tests, some use VCR, other mock in a way others in another away, I need help on that.

Regarding the changes I propose a different structure.

According to the thor docs(http://whatisthor.com/) if I want to implement subcommands like (trollolo get SUBCOMMANDS), I need to create another class that inherits from Thor and then point that class from the parent class and wrap everything around the same module.

So basically what I did was, I created a module name TrolloloCLI on `cli.rb`, then I created a dir named cli and created a class inside named Get, I copied to this new class all the get related commands and renamed it to what was defined on #15.

On the `cli.rb` file I require the `cli/get.rb`, also I had to call some methods that were private on the class Cli because of that I changed these methods to be module methods and the @@settings to be a variable of the module, this way I can call them from the new class Get and in the future from the others classes that I intend to do.

What do you guys think? Is this a good way to do the restructure? And about the tests how can I understand them fully and maybe cleanup the test suite?  (seems that @manno is working on that #193 PR)


cornelius:
I would keep it simple for now and not introduce a new namespace or rename the command class. You could just add a new class `CliGet` where you define the `get` subcommand. This can be included in the same way as all other classes in `lib/trollolo.rb` and the file could also just live in `lib/cli_get.rb`.

Calling private methods from outside of a class is usually not a great idea. They are private for a reason after all. Maybe you can find a better way?

The existing `@@settings` variable also is not a great idea. This kind of global state tends to complicate things. This might be a bigger change but I would try to move the code base into a direction where this is not needed anymore but settings are passed explicitly to the classes which need them.

Regarding the tests I would not use VCR anymore. #193 will remove them. For the command line handling it might be best to not rely on any board data and only test that the logic of handling the commands works properly.

Ana06:
@matheussbernardo regarding the `@@setings`, I agree with @cornelius that it would be better to move the code base into a direction where this is not needed anymore but settings are passed explicitly to the classes which need them.

Making `CliGet` a subclass of `Cli` may be an option if many of the methods there are needed. So that the namespace is not needed. @cornelius what do you think about that?

matheussbernardo:
Actually @Ana06  I dont think inherit from Cli makes sense because the Cli class has public methods that I do not want to inherit, and the thor documentation uses the subcommands as subclasses from Thor.

matheussbernardo:
I thought about introducing a new module that  implements

`def board_id(id_or_alias)`
`def require_trello_credentials`
`def process_global_options(options)`
`def self.settings=(s)`
and getters and setters for settings, 
then  this module is `include`in the classes that need to use them.

Or maybe use ruby open classes and include the new module on the Thor class so that all the subcommands and Cli get the methods that is needed

Ana06:
[Restructure commands](https://trello.com/c/h1OO818y/10-restructure-commands)

matheussbernardo:
@cornelius @Ana06 What do you guys think about the module?

Ana06:
@matheussbernardo the idea sounds go to me, @cornelius what do you think?

cornelius:
Yes, a module sounds good. Give it a try.

lib/cli.rb
```diff
@@ -408,7 +57,229 @@ def require_trello_credentials
   # Returns the board_id using id_or_alias. If id_or_alias matches a mapping
   # from trollolorc then the mapped id is returned or else the id_or_alias
   # is returned.
-  def board_id(id_or_alias)
+  def self.board_id(id_or_alias)
```
Ana06: this method was private before and now it is public :thinking:

lib/cli.rb
```diff
@@ -15,376 +15,25 @@
...
+  def self.process_global_options(options)
     @@settings.verbose = options[:verbose]
     @@settings.raw = options[:raw]
   end
 
-  def require_trello_credentials
+  def self.require_trello_credentials
     write_back = false
-
+
```
Ana06: extra spaces that can be removed :wink:

lib/cli.rb
```diff
@@ -15,376 +15,25 @@
...
-  private
-
-  def process_global_options(options)
+  def self.process_global_options(options)
     @@settings.verbose = options[:verbose]
     @@settings.raw = options[:raw]
   end
 
-  def require_trello_credentials
+  def self.require_trello_credentials
```
Ana06: this method was private before and now it is public :thinking:

lib/cli.rb
```diff
@@ -408,7 +57,229 @@ def require_trello_credentials
...
+
+
+  class Cli < Thor
+
+    default_task :global
+  
+    class_option :version, type: :boolean, desc: 'Show version'
+    class_option :verbose, type: :boolean, desc: 'Verbose mode'
+    class_option :raw, type: :boolean, desc: 'Raw mode'
+    # class_option 'board-id', type: :string, desc: 'id of Trello board'
```
Ana06: why is this comment here?

lib/cli.rb
```diff
@@ -408,7 +57,229 @@ def require_trello_credentials
...
+
```
Ana06: too many empty lines :see_no_evil:

lib/cli.rb
```diff
@@ -17,398 +17,36 @@
...
 
   class_option :version, type: :boolean, desc: 'Show version'
   class_option :verbose, type: :boolean, desc: 'Verbose mode'
   class_option :raw, type: :boolean, desc: 'Raw mode'
-  class_option 'board-id', type: :string, desc: 'id of Trello board'
-
-  def self.settings=(s)
-    @@settings = s
-  end
+  # class_option 'board-id', type: :string, desc: 'id of Trello board'
```
cornelius: Don't leave in code which has been commented out.

cornelius:
The code in general looks fine now but please rebase as Ana suggested so we can also see the CI run.

.rubocop_todo.yml
```diff
@@ -96,6 +96,7 @@ Layout/ExtraSpacing:
 Layout/IndentHeredoc:
   Exclude:
     - 'lib/cli.rb'
+    - 'spec/unit/cli/*.rb'
```
Ana06: you can do this automatically by running: `rubocop --auto-gen-conf`

cornelius:
looks good to me. please merge when the ci is green.

matheussbernardo:
Ready to merge!

Issue #65, opened 2017-02-26
Read burndown configuration from Trello board


cornelius:
In order to get rid of the local state in the `burndown.yaml` file there should be an option to store this data on the Trello board it belongs to. To have a smooth transition this feature should explicitly be requested by the user with a command line option. So when running `trollolo burndown --board-id=<some-board-id> --remote-state" the behavior changes to keeping the state remotely on the board. This means in particular:

* The burndown YAML is taken from the description of the burndown chart card on the specified board.
   * The burndown chart card is the first card named "burndown chart" (case insensitive match) in the first "Done" column (same card as defined in #64)
  * The YAML is embedded in the description in a code block delimited by a ` ```yaml` marker at the start and a ` ``` ` marker at the end
  * If there is more than one yaml block in the description the first one is taken
  * Any other content around the yaml block in the description is ignored by the burndown code
  * If there is no YAML configuration found the command exits with an error describing where it has looked for the information
* The burndown data is updated from the data on the Trello board and a png is generated
* The updated data is written back to the same YAML block it was taken from
* The graphics is uploaded to the burndown data card (as described in #64). When the `--remote-state` option is given, this is done unconditionally, even when no `--upload` option is explicitly given.
* When an error occurs during burndown generation or upload of the updated data it is reported by adding a comment to the burndown chart card and by sending an email to the email address configured in the YAML data (we need to add a new key for that)

PR #202, opened 2018-07-11
[WIP] Reading and write burndown from trello card

matheussbernardo:
Hello @cornelius  @Ana06 

I started to work on #65. 

It is still early work but I want to make sure is on the right path. For purpose of testing I created a command that run update. I want to make sure mostly about two things.

1 - How do you guys think is the better way to integrate this code with the BurndownChart? It was proposed a --remote-state option is this the best way? Or maybe a new command?

2 - How the remote state will work with new sprints? A new Card?


Ana06 (commented):
@matheussbernardo what is the state of this? can you please rebase and solve the merge conflicts? :thinking:

matheussbernardo:
I dont know how to implement the plot of the burndown. Because the python script depends on the YAML file. @Ana06 @cornelius

Ana06:
@matheussbernardo 

> I dont know how to implement the plot of the burndown.

what do you exactly not know? what have you already tried?

matheussbernardo:
The problem is that the python script that generates burndown depends on the YAML file and I dont want to use the YAML file. What should I do?

matheussbernardo:
I create a temp YAML file to solve the problem with the python script but I am not sure it is the best way.

cornelius:
Creating a temporary file to pass the data to the python script is ok. But you have to make sure that this is actually is a temporary file and it doesn't overwrite any other files. Alternatively you could also adapt the python script that it takes the data from standard input then you don't need to write it out.

In any case you need to write tests for the new code.

.rubocop_todo.yml
```diff
@@ -319,27 +313,34 @@ Style/Encoding:
...
     - 'lib/cli/burndown.rb'
+    - 'lib/remote_burndown.rb'
     - 'lib/scrum/sprint_cleaner.rb'
     - 'spec/integration/integration_spec_helper.rb'
     - 'spec/unit/spec_helper.rb'
 
+# Offense count: 1
+Style/MixinUsage:
+  Exclude:
+    - 'spec/unit/burndown_plot_spec.rb'
```
Ana06: some of the offenses may be because you move code around, but some other looks new. You shouldn't introduce new offenses... :eyes:

matheussbernardo: Fixed!

Ana06 (commented):
@cornelius 

> you have to make sure that this is actually is a temporary file and it doesn't overwrite any other files.

@matheussbernardo what is the state of this? :thinking:

matheussbernardo:
@Ana06 @cornelius  I create a temp directory to the files. So everytime it runs remote it create a tempdir then it uses that dir as the workind dir both ruby and python script and then i delete the temp files and the temp dir. Also I use a headless mode on the python

Ana06 (commented):
@matheussbernardo 

> @Ana06 @cornelius I create a temp directory to the files. So everytime it runs remote it create a tempdir then it uses that dir as the workind dir both ruby and python script and then i delete the temp files and the temp dir. Also I use a headless mode on the python

but still it could be that we already have a `trollolotemp` directory which you overwrites... couldn't we check if the files exist before creating them and/or generating a random folder name?

matheussbernardo:
@Ana06 Now with the mktmpdir the name is create random, I just set a prefix that is 'trollolo'

PR #207, opened 2018-08-06
Documentation for restructure commands

matheussbernardo:
This PR improves documentation to reflect the changes on the commands.

### DIFF
--- a/man/trollolo.1.md
+++ b/man/trollolo.1.md
@@ -43,195 +43,181 @@ and cards and has functionality for extracting data for burndown charts.
 Displays help about the available trollolo commands. Can take a command as
 an argument to display detailed information about it.
 
-### burndown-init -- Initialize burndown chart
+### burndown init -- Initialize burndown chart
 
-`trollolo burndown-init --board-id=<board id> --output=<directory>`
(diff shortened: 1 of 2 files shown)


man/trollolo.1.md
```diff
@@ -43,195 +43,181 @@ and cards and has functionality for extracting data for burndown charts.
...
-### list-member-boards -- List name and id of boards for a user
-
-`trollolo list-member-boards --member-id=<member id>`
-
-Get a list of all boards the given user is a member of.
-
-### set-cover -- Set picture as cover
+### set cover -- Set picture as cover
 
 `trollolo set-cover <filename> --card-id=<card id>`
```
cornelius: The dash needs to go here as well.

matheussbernardo:
Ready to merge!

Ana06:
@cornelius 

> I'm actually wondering how helpful the man page is. Seems like it is something which is not easy to maintain. Maybe it would be better to make sure that the command line help is accurate and has enough info. But I guess this is a different discussion.

I agree, now we have to maintain two docus. Should we open an issue for that? :thinking:

